PR #103, opened 2014-07-08
Remove bitenumerate; use itr macro from Iterators instead

simonster:
Can be merged once JuliaLang/Iterators.jl#33 makes it into a release


ararslan:
Is this still something we want? Any idea how performance compares to just using `enumerate`?

PR #108, opened 2014-07-29
Replace array with Base.convert

johnmyleswhite:
This pull request deprecates the `array` family of functions and replaces them with new methods for `Base.convert`. It maintains all of the same functionality, while bringing us closer into sync with Base. A similar pull request for DataFrames should be ready tonight.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -297,11 +297,18 @@ end
 #' @examples
 #'
 #' dv = @data [1, 2, 3, 4]
-#' v = array(dv)
+#' v = convert(Vector, dv)
 #'
 #' dm = @data [1 2; 3 4]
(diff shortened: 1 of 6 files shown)


yesimon:
This pull request also has the benefit of fixing infinite recursion when converting an `Array{Any,1}` with `NA` values.

```
julia> array([@data(convert(Array{Any,1}, [1:4])), NA], 0)
ERROR: interrupt
 in array at none:2 (repeats 79996 times)
```

johnmyleswhite:
@yesimon, would be great if you took the lead and rebased this change to get it working again and finished any TODO items.

ararslan:
Appears to have been superseded by #150.

PR #150, opened 2015-05-18
Replace array with Base.convert

yesimon:
This is just a rebased change from #108. Moved `array()` stuff into deprecated but otherwise didn't change the PR.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -328,40 +328,6 @@ end
 
 #' @description
 #'
-#' Turn a DataArray into an Array. Raises an error if NA's are encountered.
-#'
-#' @param da::DataArray{T} DataArray that will be converted to an Array.
-#'
-#' @returns a::Array{T} Array containing values of `da`.
-#'
-#' @examples
-#'
-#' dv = @data [1, 2, 3, 4]
-#' v = convert(Vector, dv)
-#'
-#' dm = @data [1 2; 3 4]
-#' m = convert(Matrix, dm)
-function array{T}(da::DataArray{T}) # -> Array{T}
-    Base.depwarn(
-        """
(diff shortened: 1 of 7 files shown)


johnmyleswhite:
+1

simonster:
It looks like the deprecated functions are still in `dataarray.jl` and `pooleddataarray.jl`. Maybe you missed a commit? Other than that, I think this should be good to go.

yesimon:
Fixed

simonster:
Thanks!

garborg:
Fine if I tag this, so I can start using this with DataFrames (and remove warnings from internal DataFrames code)?

simonster:
Fine with me.

garborg:
Thanks. Tagged and released.

garborg:
Thoughts?:

```julia
julia> using DataArrays

julia> dv = @data([NA, 3, 2, 5, 4])
5-element DataArrays.DataArray{Int64,1}:
  NA
 3
 2
 5
 4

julia> convert(Array, dv, 11)
5-element Array{Int64,1}:
 11
  3
  2
  5
  4

julia> convert(Matrix, dv, 11)
ERROR: MethodError: `convert` has no method matching convert(::Type{Array{T,2}}, ::DataArrays.DataArray{Int64,1}, ::Int64)
This may have arisen from a call to the constructor Array{T,2}(...),
since type constructors fall back to convert methods.
```

simonster:
Maybe I'm missing the point, but since `convert(Matrix, ::Vector)` doesn't work, I wouldn't expect `convert(Matrix, ::DataVector)` to work either.

garborg:
You're right, I thought it did but didn't check.

PR #152, opened 2015-06-04
Internal API changes due to 10525

mbauman:
There were a few internal APIs that I changed in JuliaLang/julia#10525.  This is a start at fixing some of your uses of them.  ~~I think I got everything except a few assumptions about how Base.checkbounds will work with NAType, so this fails in a few places where you expect checkbounds to throw an NAException -- it is now a MethodError.~~

There is a lot more that you could do to adjust to a post-10525 world.  Let me know if you have any questions along the way.

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -144,7 +144,7 @@ end
 end
 
 function _getindex{T}(A::DataArray{T}, I::@compat Tuple{Vararg{Union(Int,AbstractVector)}})
-    shape = Base.index_shape(I...)
+    shape = Base.index_shape(A, I...)
     _getindex!(DataArray(Array(T, shape), falses(shape)), A, I...)
 end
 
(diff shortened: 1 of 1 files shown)


mbauman:
Actually, these should be accompanied with a Compat entry to allow things to continue to work on 0.3.

mbauman:
Ah, got it.  Now the only remaining question is how to support these usages in 0.3.  (Tsk tsk for using internal APIs).  :)

mbauman:
I'm not sure that there's a sensible way to use Compat here since these aren't exported methods.  And there were only two other packages that got tripped up on this (that weren't cascading failures).

CC: @davidanthoff for ExcelReaders.jl, @kbarbary for FITSIO.jl (FITSIO might be just the change to checkbounds)

This PR has the information you'll need to get your packages up and running again after Julialang/julia#10525 changed some internal APIs regarding indexing.

kbarbary:
@mbauman Thanks for the heads up, glad to see 10525 merged!

alyst:
Oh, good that I have found this PR! I was almost desperate because of the cryptic failure in DataFrames (JuliaStats/DataFrames.jl#810). It would be nice to see this merged soon.

axsk:
Indeed #10525 broke my use of DataFrames completely. Thank you for this PR :+1:, hope it will be merged soon

nalimilan:
Any reason not to merge? The reduction in coverage appears to be due to the parts conditional on the Julia version (as expected).

simonster:
Thanks! Sorry for the slow merge.

PR #155, opened 2015-06-17
repeat() methods for DataArray and PooledDataArray

alyst:
Add repeat() methods that follow the logic of Base.repeat() (and use it internally) + tests


nalimilan:
Thanks, this is particularly useful now that `rep` is deprecated. I've rebased and fixed the spacing at https://github.com/JuliaStats/DataArrays.jl/pull/194 (please follow the spacing conventions used in the rest of the file).

PR #160, opened 2015-07-30
Import operators to avoid deprecation warnings on 0.4-dev

samuelpowell:


simonster:
Since we were already doing this in most places, I avoided the warnings by qualifying the definitions (e.g. `Base.(:+)` instead of `+`). But thanks for looking into this!

Issue #105, opened 2014-07-10
Keep index of pool for PooledDataArray?


simonster:
At the moment, `setindex!` has to look up every value in the pool, so the computational complexity of setting `n` indices in a PooledDataArray with a pool of size `m` is O(mn). If we kept a `Dict{eltype(pda),Int}` that maps values to indices in the pool, we could reduce this to O(n).

PR #161, opened 2015-08-05
Speed up permute!(PooledDataArray)

alyst:
Avoid slow `setindex!(PooledDataArray)` (see #105) by permuting `.refs` only. This should speed up data frame sorting.

PR redefines `Base.permute!!()`, which is not exported by `Base`. The reason is that `permute!(v, p)` calls `permute!!(v, copy(p))`, but `DataFrames` uses `permute!!()` directly to avoid unnecessary temporary array allocation for each column.

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -443,6 +443,16 @@ reorder(x::PooledDataArray, y::AbstractVector...) = reorder(mean, x, y...)
 
 Base.reverse(x::PooledDataArray) = PooledDataArray(RefArray(reverse(x.refs)), x.pool)
 
+function Base.permute!!{T<:Integer}(x::PooledDataArray, p::AbstractVector{T})
+    Base.permute!!(x.refs, p)
+    x
+end
+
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. Adding a method to an unexported function isn't great, but as long as `DataFrames` uses it... Maybe it should be exported?

src/DataArrays.jl
```diff
@@ -52,6 +52,8 @@ module DataArrays
            xtab,
            xtabs
 
+import Base: permute!!, ipermute!!
```
nalimilan: You don't need this line: as you can see, the package doesn't import any functions it uses.

nalimilan:
Ah, and could you add tests for the new functions?

src/DataArrays.jl
alyst: Thanks. I suspected that it's not required, but my understanding of `import` is rather vague. Could you please clarify in what case `import` is really needed?

nalimilan: You only need it if you want to write `permute!! = ...` instead of `Base.permute!! = ...` .

alyst:
I would merge it soon, if no one objects.

PR #162, opened 2015-08-24
Turn on precompilation

timholy:
Fixes https://github.com/dcjones/Gadfly.jl/pull/667#issuecomment-133991074

### DIFF
--- a/src/DataArrays.jl
+++ b/src/DataArrays.jl
@@ -1,3 +1,5 @@
+VERSION >= v"0.4.0-dev+6641" && __precompile__()
+
 module DataArrays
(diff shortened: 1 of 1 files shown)


timholy:
Thanks to @stevengj.

johnmyleswhite:
Thanks. This should work fine since we've been doing it already to precompile DataFrames (and because DataArrays shouldn't be doing anything too wacky).

timholy:
Needs to be tagged if we are to avoid Travis failures in Gadfly like https://travis-ci.org/dcjones/Gadfly.jl/jobs/77129510

CC @dcjones

PR #163, opened 2015-08-26
isna() for scalar case

alyst:
It significantly reduces memory allocation overhead in `DataFrames` for me on 0.4.


johnmyleswhite:
It seems like we're probably being too liberal with indexing if this is needed since I would have thought the other cases would handle this.

simonster:
Most likely there is an issue with Compat's `@nsplat`. In any case once we can drop 0.3 support we should replace as much of the Cartesian use as possible with generic functions.

PR #165, opened 2015-09-04
Add symbol method for NA.

Ismael-VC:
I needed to define this for a case like this:

``` julia
julia> using DataFrames

julia> df = DataFrame(Test = @data ["a", NA, "c"])
3x1 DataFrame
| Row | Test |
|-----|------|
| 1   | "a"  |
| 2   | NA   |
| 3   | "c"  |

julia> test = countmap(df[:Test])
Dict{Union(NAtype,ASCIIString),Int64} with 3 entries:
  "c" => 1
  NA  => 1
  "a" => 1

julia> test_count = DataFrame(;[symbol(k) => v for (k, v) in test]...)
ERROR: `symbol` has no method matching symbol(::NAtype)
 in anonymous at no file

julia> Base.symbol(x::NAtype) = :NA
symbol (generic function with 7 methods)

julia> test_count = DataFrame(;[symbol(k) => v for (k, v) in test]...)
1x3 DataFrame
| Row | NA | c | a |
|-----|----|---|---|
| 1   | 1  | 1 | 1 |
```

### DIFF
--- a/src/natype.jl
+++ b/src/natype.jl
@@ -20,6 +20,7 @@ end
 const NA = NAtype()
 
 Base.show(io::IO, x::NAtype) = print(io, "NA")
+Base.symbol(x::NAtype) = :NA
 
 type NAException <: Exception
     msg::String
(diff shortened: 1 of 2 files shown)


Ismael-VC:
I have added a test, I don't know why does coverage drops down with this commit. :worried:

johnmyleswhite:
Don't worry about the coverage thing.

I'm ok with this change, but the use case seems pretty problematic. Are you really sure you want to be doing the things you're showing as examples?

PR #169, opened 2015-10-04
Fix 0.4 deprecations

hayd:
Union(..) -> Union{..}, String -> AbstractString, FloatingPoint -> AbstractFloat

The two remaining messages are:

1. cartesianmap
2. {..} syntax (Note: replacing with Any[..] fails

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -5,7 +5,7 @@ using Base.Broadcast: bitcache_chunks, bitcache_size, dumpbitcache,
 
 # Check that all arguments are broadcast compatible with shape
 # Differs from Base in that we check for exact matches
-function check_broadcast_shape(shape::Dims, As::Union(AbstractArray,Number)...)
+function check_broadcast_shape(shape::Dims, As::(@compat Union{AbstractArray,Number})...)
     samesize = true
     for A in As
         if ndims(A) > length(shape)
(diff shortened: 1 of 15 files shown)


hayd:
Hmm shoulda checked on 0.3 before, the error appears to be due to the following:
```
julia> @compat Union{Array{Int}, Void}
Array{Int64,N}

julia> @compat Union{Array{Int}, Nothing}
Union(Array{Int64,N},Nothing)

julia> isa(nothing, Void)
false
```

simonster:
I think you're on an old Compat. I get `@compat(Union{Array{Int}, Void}) == Union{Array{Int},Nothing}` as expected on 0.3, but all other uses of `Void` (e.g. in `isa`) should also have `@compat` in front of them. (`Void` is `None == Union{}` on 0.3 is `typeof(nothing) == Nothing` on 0.4.)

hayd:
Tests appear to pass on both now (so I've squashed).

There is this warning, which I don't see on my machine. But I suspect that isn't new...

    WARNING: could not attach metadata for @simd loop.

johnmyleswhite:
I think this is an improvement over the status quo, so I'm merging it. We can try to deal with the `@simd` issue in subsequent work.

hayd:
Thanks! Would be great to push a release soon.

I might try and do the same for DataFrames... it would be great if there was a tool which just did the rewrite inplace! For now, it's vim macros all the way.

matthieugomez:
I think the @simd issue just comes from the fact [inlining is turned off](https://groups.google.com/forum/#!topic/julia-dev/u6G7f9mxaHI)

PR #172, opened 2015-10-20
An uninitialized DataArray is now convertable.

gustafsson:
Simple test that previously caused an error:
> convert (DataArray{Integer,1},DataArray(Any,2))

That might seem pointless but consider this:
> a = DataArray(Any,2)
> a[1] = 2;
> a[2] = NA;
> convert(DataArray{Integer,1},a)
All values are defined, but the conversion still failed. This commit fixes this.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -583,7 +583,11 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
 #' dv_alt = convert(DataVector{Float64}, dv)
 function Base.convert{S, T, N}(::Type{DataArray{S, N}},
                                x::DataArray{T, N}) # -> DataArray{S, N}
-    return DataArray(convert(Array{S}, x.data), x.na)
+    v = Array{S,N}(size(x.data)...)
+    @inbounds for i = find(!x.na)
+        v[i] = convert(S, x.data[i])
+    end
+    return DataArray(v, x.na)
 end
 
 #' @description
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -584,8 +584,10 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
 function Base.convert{S, T, N}(::Type{DataArray{S, N}},
                                x::DataArray{T, N}) # -> DataArray{S, N}
     v = similar(x.data, S)
-    @inbounds for i = find(!x.na)
-        v[i] = convert(S, x.data[i])
+    @inbounds for i = 1:length(x)
+        if !x.na[i]
+            v[i] = convert(S, x.data[i])
+        end
     end
     return DataArray(v, x.na)
 end
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -583,7 +583,7 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
 #' dv_alt = convert(DataVector{Float64}, dv)
 function Base.convert{S, T, N}(::Type{DataArray{S, N}},
                                x::DataArray{T, N}) # -> DataArray{S, N}
-    v = Array{S,N}(size(x.data)...)
+    v = similar(x.data, S)
     @inbounds for i = find(!x.na)
         v[i] = convert(S, x.data[i])
     end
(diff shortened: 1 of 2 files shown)


gustafsson:
This is still an issue in master. Any comments on what would be needed in order to merge this?

src/dataarray.jl
```diff
@@ -569,7 +569,11 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
 #' dv_alt = convert(DataVector{Float64}, dv)
 function Base.convert{S, T, N}(::Type{DataArray{S, N}},
                                x::DataArray{T, N}) # -> DataArray{S, N}
-    return DataArray(convert(Array{S}, x.data), x.na)
+    v = Array(S, size(x.data))
+    for i = find(!x.na)
+        v[i] = convert(S, x.data[i])
+    end
```
ararslan:
```julia
v = Array{S,N}(size(x.data))
@inbounds for i = find(!x.na)
    v[i] = convert(S, x.data[i])
end
```
Might have to splat the `size`, can't recall offhand

ararslan:
This is a rather old PR so I'd recommend rebasing from master just in case

gustafsson:
@ararslan thanks, I implemented your feedback

ararslan:
Looks like the CI tests are still failing. Since the tests are currently passing on master I assume the changes in this PR are the cause, so that should be addressed before this is merged.

src/dataarray.jl
```diff
@@ -583,7 +583,11 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
 #' dv_alt = convert(DataVector{Float64}, dv)
 function Base.convert{S, T, N}(::Type{DataArray{S, N}},
                                x::DataArray{T, N}) # -> DataArray{S, N}
-    return DataArray(convert(Array{S}, x.data), x.na)
+    v = Array{S,N}(size(x.data)...)
```
simonster: This should actually be `Array{S}(size(x.data))` or just `similar(x.data, S)`

ararslan: I figured the `Array{S,N}` would help performance since `Array{S,N}` is a leaf type whereas `Array{S}` isn't. :+1: for `similar` though.

ararslan: I did some brief benchmarking and it looks like `Array{S,N}(size(x.data))` is the fastest among `Array{S,N}(t)`, `Array{S}(t)`, and `similar(v, S)` (for `v` an array and `t` a tuple of its dimensions).

simonster: AFAICT, `Array{S,N}` works if the dimensions aren't splatted on Julia 0.5, but not Julia 0.4.

It shouldn't matter which variant you choose here. Julia 0.5 appears to generate identical code for `Array{S,N}(size(x))` and `Array{S}(size(x))`. `similar(x, S)` looks like it does some extra checks on the array dimensions, but that results effectively zero overhead relative to the cost of actually allocating an Array.

ararslan: > AFAICT, `Array{S,N}` works if the dimensions aren't splatted on Julia 0.5, but not Julia 0.4.

:disappointed: 

> It shouldn't matter which variant you choose here. Julia 0.5 appears to generate identical code for `Array{S,N}(size(x))` and `Array{S}(size(x))`.

Interesting, that's good to know. Thanks!

> `similar(x, S)` looks like it does some extra checks on the array dimensions, but that should have effectively zero overhead relative to the cost of actually allocating an Array.

Cool, sounds like `similar` is the best choice then.

gustafsson: `Array{S,N}(size(x.data))` doesn't exist (I'm running Julia 0.4.6). And `Array{S,N}(size(x.data)...)` only works for `N in [1,2]`. I'm pushing a commit that uses similar now, I think it expresses the intent well. If it isn't as fast as the fastest alternative we should be able to instead improve on how `similar` works since it has enough information in compile time to be just as fast in this case right?

src/dataarray.jl
```diff
@@ -583,7 +583,11 @@ Base.convert{T, N}(::Type{DataArray}, x::AbstractArray{T, N}) =
...
+    @inbounds for i = find(!x.na)
```
nalimilan: This quite inefficient. Better go over all elements in `x`, and add an `if !x.na[i]` inside the loop.

nalimilan:
Thanks!

PR #173, opened 2015-10-30
Fix breakage on 0.5-dev.

yuyichao:
I'm not sure how much of these can go to `Compat.jl` so I just implement everything here....

c.c. @timholy 

Ref JuliaLang/julia#13803

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -1,7 +1,20 @@
 using DataArrays, Base.@get!
 using Base.Broadcast: bitcache_chunks, bitcache_size, dumpbitcache,
-                      promote_eltype, broadcast_shape, eltype_plus, type_minus, type_div,
-                      type_pow
+                      promote_eltype, broadcast_shape, eltype_plus
+
+if isdefined(Base.Broadcast, :type_minus)
+    using Base.Broadcast: type_minus, type_div, type_pow
+    const _type_minus = type_minus
+    const _type_rdiv = type_div
+    const _type_ldiv = type_div
+    const _type_pow = type_pow
+else
+    using Base.Broadcast: promote_op
+    _type_minus(T, S) = promote_op(Base.SubFun(), T, S)
+    _type_rdiv(T, S) = promote_op(Base.RDivFun(), T, S)
+    _type_ldiv(T, S) = promote_op(Base.LDivFun(), T, S)
+    _type_pow(T, S) = promote_op(Base.PowFun(), T, S)
+end
 
 # Check that all arguments are broadcast compatible with shape
 # Differs from Base in that we check for exact matches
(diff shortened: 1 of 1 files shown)


timholy:
I'm not sure those are material for Compat anyway since they were unexported functions. But if others feel differently, I'm fine with wherever they end up.

yuyichao:
Oh SegFault, great .....

Also note that if we backport JuliaLang/julia#13803, the version check in this PR will needs to be tweaked....

yuyichao:
Hmm. I couldn't reproduce the segfault but I'm getting some very strange errors in `reducedim.jl` test.

Depending on what I do before, I get either
```jl
ERROR: LoadError: UndefVarError: N not defined
 in _mapreducedim! at ./no file:4294967295
```
or
```jl
ERROR: LoadError: MethodError: `_nextract` has no method matching _nextract(::Type{Any}, ::Symbol, ::Expr)
Closest candidates are:
  _nextract(::Int64, ::Symbol, ::Expr)
  _nextract(::Int64, ::Symbol, ::Symbol)
 in func_for_method_checked at ./reflection.jl:248
```

when I try to call `@code_warntype Base._mapreducedim!(+, Base.AndFun(), Bool[], BitArray(10))`

This only happens if the module is precompiled but not if I add `--compilecache=no` or define the corresponding method at runtime again.

Can anyone reproduce this?

@vtjnash The method that is causing this issue directly is [here](https://github.com/JuliaStats/DataArrays.jl/blob/ef0ee793fa7c20eeb5ba7a2d0ebbc3130ea1f3ff/src/reducedim.jl#L10). Is there anything in it that looks suspicious for you in terms of precompilation compatibility? Also, is there a way I can check the `.ji` file and see if there's anything strange in there?

yuyichao:
The first error `UndefVarError(:N)` is what I get when running the test locally without `--compilecache=no`.

yuyichao:
Force pushed to trigger a travis rerun.

yuyichao:
.... this time the error is the same one I see locally .....

https://travis-ci.org/JuliaStats/DataArrays.jl/jobs/88360029

yuyichao:
The precompilation error is caused by https://github.com/JuliaLang/julia/pull/13677 ....

src/broadcast.jl
```diff
@@ -1,7 +1,20 @@
 using DataArrays, Base.@get!
 using Base.Broadcast: bitcache_chunks, bitcache_size, dumpbitcache,
-                      promote_eltype, broadcast_shape, eltype_plus, type_minus, type_div,
-                      type_pow
+                      promote_eltype, broadcast_shape, eltype_plus
+
+if VERSION < v"0.5.0-dev+1016"
```
tkelman: can you use an `isdefined` check here rather than a version comparison?

yuyichao: Good point, will do when I'm back.

yuyichao:
Updated the check to use `isdefined` so that it should work on 0.4 if we end up backporting JuliaLang/julia#13803 .

The nightly CI will fail until commit https://github.com/JuliaLang/julia/commit/5498570e2cb6a9b7065b7dc592c076b9b3125962 is included in the nightly binary (hopefully tomorrow...)

tkelman:
should be up already: http://buildbot.e.ip.saba.us:8010/builders/package_tarball64/builds/145, hit restart if the `versioninfo()` looks too hold

yuyichao:
You are right, the travis run fetch exactly the commit that has the fix. This should be ready to go then.

tkelman:
@simonster et al, okay to merge this?

tkelman:
Note that this needs to be tagged before https://github.com/JuliaLang/julia/pull/13803 can be considered for backporting. Not urgent as it's almost certainly going to be for 0.4.2, not 0.4.1.

PR #176, opened 2015-12-18
Allow replacing with results of a function

paulvictor:
Create another type which can replace with results of evaluating a function, optionally providing the data array itself as an argument. Use case: Replace with samples drawn from the distribution of non NA values.

### DIFF
--- a/src/DataArrays.jl
+++ b/src/DataArrays.jl
@@ -25,6 +25,7 @@ module DataArrays
            each_failna,
            each_dropna,
            each_replacena,
+           each_replacenawithfunctionresult,
            EachFailNA,
            EachDropNA,
            EachReplaceNA,
--- a/src/abstractdataarray.jl
+++ b/src/abstractdataarray.jl
@@ -168,3 +168,19 @@ function Base.next(itr::EachReplaceNA, ind::Integer)
     item = isna(itr.da, ind) ? itr.replacement : itr.da[ind]
     (item, ind + 1)
 end
+
+type EachReplaceNAWithFunctionResult{T}
+    da::AbstractDataArray{T}
+    f::Function
+    withData::Bool
+end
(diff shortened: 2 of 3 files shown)


nalimilan:
Thanks, but I'm not sure a dedicated function and type are really needed here. I find an explicit loop clearer, and almost as short as your version:
```julia
[isna(dv, i) ? mean(dropna(dv)) : dv[i] for i in eachindex(dv)]
```
or the in-place version
```julia
for i in eachindex(dv)
    if isna(dv, i) dv[i] = m end
end
``

paulvictor:
True, but that's the case with each_replacena too, here the example given seems simple, but given that a function can be used opens up a number of use cases, like replacing NA's with values drawn from a distribution described by the non-NA values(which need not be the same for each invocation), here again, this can be replaced by the transformation as you described, but I find this style better, since you make it more declarative(by passing the function(the what)) than imperative(generating and assigning the values(the how)). On a note aside, there are tests failing on travis and the coverage has been reported to be down, but no test errors occur when I build locally, is there anything I'm missing? Attaching a screenshot of how the tests ran, please give your feedback on the same.

screenshot 2015-12-19 22 25 26

nalimilan:
Others should tell whether they find it useful before you do more work on it, but I think you should at least merge `each_replacenawithfunctionresult` into `each_replacena`. You could simply add a method for `Callable` arguments.

I'm not sure why the tests are failing at the moment.

paulvictor:
@nalimilan , Can you check https://github.com/paulvictor/DataArrays.jl/commit/365f1a34235bd79e8b085170082e516abf328cdf, if that's what you meant? I can provide a squashed commit later if the code makes sense.

nalimilan:
Yeah, something like that, but as I said I'll let other comment.

paulvictor:
Thanks @nalimilan , will wait too...

PR #175, opened 2015-11-29
enable AppVeyor

tkelman:
I think there may be a test failure on Windows at the moment - can someone turn AppVeyor on for this repo on their account? For people with commit access the AppVeyor site should handle enabling the webhook, but if it doesn't work then I can dig up specific instructions.

PR #185, opened 2016-05-11
Make DataArrays ready for 0.5

andreasnoack:
All kind of stuff: new indexing, new iterations, no functors, new `getfield` syntax for operators, string changes and some other deprecations. Tests are passing now on master but the string story is not completely settled yet in `Compat.jl` so we'll need a new tag there before tests pass on 0.4. It might also be the right time to stop 0.3 support. I don't want to spend much time on making the tests pass there.

I should also mention that some of the tests are now slower than they are on 0.4. However, it's difficult to say if this is because of my changes here or because of the general slowdown on master. I profiled the `reducedim` tests and there is spent a lot of time in `inference` (even after the first run) and I can figure out why.

### DIFF
--- a/src/DataArrays.jl
+++ b/src/DataArrays.jl
@@ -1,4 +1,4 @@
-VERSION >= v"0.4.0-dev+6641" && __precompile__()
+__precompile__()
 
 module DataArrays
     using Base.Cartesian, Compat, Reexport
--- a/src/DataArrays.jl
+++ b/src/DataArrays.jl
@@ -2,9 +2,13 @@ VERSION >= v"0.4.0-dev+6641" && __precompile__()
 
 module DataArrays
     using Base.Cartesian, Compat, Reexport
+    import Compat.String
     @reexport using StatsBase
 
     const DEFAULT_POOLED_REF_TYPE = UInt32
+
+    import Base: ==, !=, >, <, >=, <=, +, -, *, !, &, |, $, ^, /,
+      .==, .!=, .>, .<, .>=, .<=, .+, .-, .*, .%, ./, .\, .^
 
     export @data,
            @pdata,
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -10,10 +10,10 @@ if isdefined(Base.Broadcast, :type_minus)
     const _type_pow = type_pow
 else
     using Base.Broadcast: promote_op
-    _type_minus(T, S) = promote_op(Base.SubFun(), T, S)
-    _type_rdiv(T, S) = promote_op(Base.RDivFun(), T, S)
-    _type_ldiv(T, S) = promote_op(Base.LDivFun(), T, S)
-    _type_pow(T, S) = promote_op(Base.PowFun(), T, S)
+    _type_minus(T, S) = promote_op(@functorize(-), T, S)
+    _type_rdiv(T, S) = promote_op(@functorize(/), T, S)
+    _type_ldiv(T, S) = promote_op(@functorize(\), T, S)
+    _type_pow(T, S) = promote_op(@functorize(^), T, S)
 end
 
 # Check that all arguments are broadcast compatible with shape
@@ -38,10 +38,10 @@ end
 # Get ref for value for a PooledDataArray, adding to the pool if
 # necessary
 _unsafe_pdaref!(Bpool, Brefdict::Dict, val::NAtype) = 0
-function _unsafe_pdaref!{K,V}(Bpool, Brefdict::Dict{K,V}, val)
+function _unsafe_pdaref!(Bpool, Brefdict::Dict, val)
     @get! Brefdict val begin
         push!(Bpool, val)
-        convert(V, length(Bpool))
+        length(Bpool)
     end
 end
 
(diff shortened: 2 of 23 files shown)


src/broadcast.jl
```diff
@@ -208,8 +208,8 @@ for bsig in (DataArray, PooledDataArray), asig in ((@compat Union{Array,BitArray
             B
         end
         # ambiguity
-        Base.map!(f::Base.Callable, B::$bsig, r::Range) =
-            invoke(Base.map!, (Base.Callable, $bsig, $asig), f, B, r)
+        # Base.map!(f::Base.Callable, B::$bsig, r::Range) =
```
nalimilan: If this isn't needed, why not remove it completely?

src/DataArrays.jl
```diff
@@ -2,10 +2,14 @@ VERSION >= v"0.4.0-dev+6641" && __precompile__()
 
 module DataArrays
     using Base.Cartesian, Compat, Reexport
+    import Compat.String
     @reexport using StatsBase
 
     const DEFAULT_POOLED_REF_TYPE = UInt32
 
+    import Base: ==, !=, >, <, >=, <=, +, -, *, !, &, |, $, ^, /,
```
nalimilan: Why do you import these and not e.g. `length`?

src/broadcast.jl
andreasnoack: I simply forgot that. I'll remote it.

src/DataArrays.jl
andreasnoack: I only import the operators. I actually like to use the qualified method definition instead of importing the names but I (think I) had to do it this way because of the changes to `getfield`.

src/extras.jl
```diff
@@ -98,4 +98,4 @@ function rep(x::AbstractVector; times::Integer = 1, each::Integer = 1)
     rep(x, times, each)
 end
 
-rep(x::Any, times::Integer) = fill(x, times)
+rep(x::Number, times::Integer) = fill(x, times)
```
nalimilan: Why is this needed?

src/indexing.jl
```diff
@@ -186,25 +186,31 @@ end
     end
 end
 
+if VERSION > v"0.5-"
+    Base.unsafe_getindex(x::Number, i::Int) = (@inbounds r = x[i]; r)
```
nalimilan: This looks weird. Why is is necessary?

andreasnoack: The method was removed from Base in one of the indexing overhauls but it is used by one of the indexing methods in this package. It can probably be removed when somebody has the time to update the indexing methods here systematically.

src/extras.jl
andreasnoack: It would clash with the definition at line 82. For some reason, I don't think it is detected on 0.4.

andreasnoack:
@nalimilan Thanks for the comments. I've made the necessary changes for the 0.4 tests to pass and I've removed that 0.3 support since 0.3 doesn't have `@generated`.

Anybody knows why AppVeyor testing is enabled but no `.appveyor.yml` and who's in charge of AppVeyor for JuliaStats?

src/DataArrays.jl
nalimilan: OK.

src/indexing.jl
nalimilan: Indeed, that's what I thought. Maybe a FIXME would be useful here. I'd rather work on the replacement for DataArrays than work on improving it, but at least that would make issues visible.

test/reducedim.jl
```diff
@@ -56,7 +56,7 @@ function safe_mapslices{T}(f::Function, A::AbstractArray{T}, region, skipna)
         idx[d] = 1:size(A,d)
     end
 
-    r1 = f(reshape(A[idx...], Asliceshape); skipna=skipna)
+    r1 = f(copy(reshape(A[idx...], Asliceshape)); skipna=skipna)
```
nalimilan: Is a copy really necessary, since `r1` seems to be read-only?

nalimilan:
Sorry, I was interrupted in the middle of the review. Thanks for doing this. I agree dropping 0.3 is OK at this point.

Regarding AppVeyor, looks like we stopped in the middle of enabling it. See https://github.com/JuliaStats/DataArrays.jl/pull/175.

test/reducedim.jl
andreasnoack: The reductions don't work with `ReshapedArray`s and the copy makes it a `DataArray`. The main point here, though, is that this is just a test function to make sure that the "real" reduction functions work correctly with respect to `NA`s.

andreasnoack:
The are still a lot of packages to go through before we are ready for 0.5 so I'll probably merge this soon and start to look into `DataFrames`.

johnmyleswhite:
LGTM

nalimilan:
OK to merge then?

andreasnoack:
Not yet. There are some ambiguities now that I'll have to resolve. I'm looking into it .

src/indexing.jl
tkelman: can this be removed yet?

ararslan: @tkelman It's not on master, so it has been removed at some point

tkelman: it's still there, just not version conditional any more

tkelman: https://github.com/JuliaStats/DataArrays.jl/blob/v0.4.0/src/indexing.jl#L128

ararslan: Oh yeah, I see now. https://github.com/JuliaStats/DataArrays.jl/pull/240

PR #186, opened 2016-05-13
Deprecate rep(), head() and tail()

nalimilan:
This package is not the place for utilities functions. Most cases
are supported by repeat(). rep(::AbstractVector, ::AbstractVector)
has no replacement currently, but it is a less common form and
should be moved to Julia Base or another package.

I'm trying to clean up DataFrames, and it extends this function. I figured we'd better prompt people to use Base's `repeat` ASAP. Though `repeat` does not support `AbstractArrays` until https://github.com/JuliaLang/julia/pull/14082 is merged and backported, which is why I kept the existing functions instead of calling it.

I don't know what to do about `rep(::AbstractVector, ::AbstractVector)`: should Base offer this feature via `repeat`? Or via another function? Should it go to a utilities package?


nalimilan:
Any comments?

nalimilan:
I'll merge in a few days if nobody objects.

johnmyleswhite:
+1 to this

PR #189, opened 2016-05-23
Check for NA in DataArray inner constructor

bkamins:
Currently inner constructor of DataArray does not check if `d` argument contains `NA` entries.
The proposed change overwrites values supplied in `m` argument with 'true' for elements of 'd'  that are 'NA'.
Closes https://github.com/JuliaStats/DataFrames.jl/issues/974.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -29,10 +29,14 @@ type DataArray{T, N} <: AbstractDataArray{T, N}
             throw(ArgumentError(msg))
         end
         # additionally check if d does not contain NA entries
-        for i in eachindex(d)
-            if isdefined(d, i) && isna(d, i)
-                m[i] = true
+        if eltype(d) == Any
+            for i in eachindex(d)
+                if isdefined(d, i) && isna(d, i)
+                    m[i] = true
+                end
             end
+        elseif eltype(d) <: NAtype
+            m = trues(m)
         end
         new(d, m)
     end
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -28,7 +28,8 @@ type DataArray{T, N} <: AbstractDataArray{T, N}
             msg = "Data and missingness arrays must be the same size"
             throw(ArgumentError(msg))
         end
-        new(d, m)
+        # additionally check if d does not contain NA entries
+        new(d, isna(d) | m)
     end
 end
 
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -29,7 +29,12 @@ type DataArray{T, N} <: AbstractDataArray{T, N}
             throw(ArgumentError(msg))
         end
         # additionally check if d does not contain NA entries
-        new(d, isna(d) | m)
+        for i in eachindex(d)
+            if isdefined(d, i) && isna(d, i)
+                m[i] = true
+            end
+        end
+        new(d, m)
     end
 end
 


nalimilan:
Looks like you need to fix or adapt a test...

bkamins:
I have fixed the problem with `isna(d)`. It was caused by `#undef` in arrays of `Any`.

However this gives rise to the following issues:

* `isna` does not correctly handle `#undef` values (it gives an error and should report `false` as in other places of the package `#undef` is not treated as `NA`)
* `anyna`, `allna`, `dropna` for non-`DataArray` argument do not correctly handle `NA` in arrays of `Any` or of `NAtype`.

I can open new issues for these problems but I do not want to pollute issues if they would be obsolete in near future (when replaced by `Nullable` approach).

bkamins:
I have fixed the problem with `isna(d)` that was caused by `#undef` values.

This however gives rise to the following new issues:

* `isna` does not correctly handle `#undef` values; it should return `false` and rises an error instead (in other places `DataArrays` package handles `#undef` as not-`NA`);
* functions `anyna`, `allna` and `dropna` do not correctly handle arrays of `Any` or `NAtype` that have `NA` entries (they assume that ordinary arrays do not contain `NA` values).

I can create new issues for these problems but I do not know if I should as I do not want to pollute repository with them if they would be obsolete soon (I understand that the migration to `Nullable` is implemented and it would make these changes obsolete).

nalimilan:
I don't think `isna` can do much on undefined values other than raising an error. You're not supposed to call any function on undefined values.

Also, I wonder whether the approach you chose is the best one. I think it would be more natural to do that check here: https://github.com/bkamins/DataArrays.jl/blob/28ea6d1fb07ab1d00f337e009f9c57ebe4c92eb0/src/dataarray.jl#L555 That would avoid the cost of going over the DataArray for places where we know the input is correct.

`anyna` and friends could also be fixed, but I won't encourage you to waste your time on this. If you want to help improving the data management packages in Julia, you'd better play with NullableArrays and JuliaDB packages, and try to spot bugs or missing features.

src/dataarray.jl
```diff
@@ -28,6 +28,16 @@ type DataArray{T, N} <: AbstractDataArray{T, N}
             msg = "Data and missingness arrays must be the same size"
             throw(ArgumentError(msg))
         end
+        # additionally check if d does not contain NA entries
+        if eltype(d) == Any
+            for i in eachindex(d)
+                if isdefined(d, i) && isna(d, i)
```
nalimilan: It would make sense to treat undefined elements as missing, wouldn't it? That would give:
`m[i] = !isdefined(d, i) || isna(d, i)`

bkamins: I thought it was a design decision to allow `#undef` in `DataArray` that is why I have proposed the original code. This preserves the following behavior:
```
DataArray(Array(Any, 5))
```
consists of five `#undef` values. After your change it would contain five `NA` values.
In summary - it is a design decision and I left the original behavior in such cases (I can image situations in which `#undef` would mean something different than `NA`).

nalimilan: Yeah, maybe there's a use case for that. Let's apply minimal changes for now.

nalimilan:
Could you add a test for the new behavior?

bkamins:
Here are the proposed tests. I have accidentally removed my repo and I do not know how to amend this PR, so I send the lines below.

```
    @test isequal(DataArray([NA, NA], [true, true]), DataArray([NA, NA], [false, false]))
    @test isequal(DataArray(Any[1, NA], [false, true]), DataArray(Any[1, NA], [false, false]))
    @test isequal(DataArray(Any[1, 2], [false, true]), DataArray(Any[1, NA], [false, false]))
```

@nalimilan: I hope you know how to update the PR :). Related to our discussion of `#undef` in `DataArray` there is a following problem (which might change our conclusion):
```
z = DataArray(Array(Any,2))
isdefined(z, 1) # returns true
z[1] # throws UndefRefError
```
so maybe `#undef` should be not allowed in `DataArray` as it is not handled correctly.

bkamins:
Additionally following JuliaLang/julia#12862 probably `isdefined` should be replaced by `isassigned` (still I am not able to amend it in PR)?

nalimilan:
Are you able to pull from your branch at least?

nalimilan:
Let's leave the `#undef` issue for another day. I'd rather not spent too much time on DataArrays as we already have a lot to do on `NullableArrays`.

bkamins:
I can pull but not push. I can create new PR and close this one.

nalimilan:
Yeah, will be easier.

PR #190, opened 2016-05-29
remove methods that are deprecated in Base

tkelman:
now removed in 0.5 - defining these wasn't giving deprecation warnings,
but calling them should have been?

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -590,38 +590,6 @@ end
 
 data(a::AbstractArray) = convert(DataArray, a)
 
-#' @description
-#'
-#' Convert a DataArray to an Array of int, float or bool type.
-#'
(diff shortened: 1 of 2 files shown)


tkelman:
Why would this cause a stack overflow? Someone who know this code is going to have to track that down, this is the cause of several dozen failures on pkgeval.

tkelman:
nevermind, was just a call to float which hasn't been deprecated so I shouldn't have deleted it (or the test can be changed first, will leave that to a maintainer here).

If this can be merged and tagged in the next 5 hrs it'll make it into the next pkgeval run.

sfchen:
This pr should be merged asap coz this issue is blocking DataFrames.

tkelman:
Not sure why win64 appveyor is failing to clone metadata...

PR #192, opened 2016-06-03
Use Base.promote_op() for arithmetic operators

nalimilan:
This fixes support for cases in which the return type is different
from the inputs, in particular Date. Also makes handling of /
a bit less of a special case.

Fixes https://github.com/JuliaStats/DataFrames.jl/issues/987.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -1,3 +1,11 @@
+promote_op{R,S}(f::Any, ::Type{R}, ::Type{S}) =
+    Base.promote_op(f, R, S)
+
+# Required for /(::Int, ::Int)
+if VERSION < v"0.5.0-dev"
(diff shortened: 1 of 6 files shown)


src/operators.jl
```diff
@@ -1,3 +1,11 @@
+promote_op{R,S}(f::Any, ::Type{R}, ::Type{S}) =
+    Base.promote_op(f, R, S)
+
+# Required for /(::Int, ::Int)
+if VERSION < v"0.5.0-dev"
```
tkelman: would be better to be more specific about the cutoff here

nalimilan: I didn't bother to check the exact version since the fallback works everywhere. The point of the check is just to make it obvious that we can remove this when dropping support for 0.4.

PR #195, opened 2016-06-07
Allowed bare Ints as repeat keyword args

bramtayl:
This prevents breakage of stack and crossjoin in DataFrames.jl


nalimilan:
Thanks but I fixed it separately without noticing. Sorry.

PR #202, opened 2016-07-02
add a less specific version of setlevels!; fixes #201 and tests on Gadfly.jl

### LINKED ISSUE #201: Type signature of setlevels! is too strict
This has been causing the tests on [Gadfly](https://github.com/dcjones/Gadfly.jl) master to be broken for some time

```{.julia}
julia> using RDatasets, DataFrames

julia> barley = dataset("lattice", "barley");

julia> setlevels!(barley[:Year], ["1931", "1932"])
ERROR: MethodError: `setlevels!` has no method matching setlevels!(::DataArrays.PooledDataArray{ByteString,UInt8,1}, ::Array{ASCIIString,1})
Closest candidates are:
  setlevels!{T,R}(::DataArrays.PooledDataArray{T,R,N}, ::AbstractArray{T,1})
  setlevels!{T,R}(::DataArrays.PooledDataArray{T,R,N}, ::Dict{T,T})
  setlevels!{T,R}(::DataArrays.PooledDataArray{T,R,N}, ::Dict{T,Any})

julia> setlevels!(barley[:Year], ByteString["1931", "1932"]);
```
Typing the array shouldn't be necessary

tlnagy:
Currently the type signature of setlevels! is extremely strict and it would be best to allow the user some more flexibility. This should also fix Gadfly's tests being broken (at least on 0.4).

See #201

### DIFF
--- a/test/pooleddataarray.jl
+++ b/test/pooleddataarray.jl
@@ -27,6 +27,8 @@ module TestPDA
     @assert levels(setlevels!(copy(p), [1,8,9, 10])) == [1, 8, 9, 10]
     @assert levels(setlevels!(copy(p), Dict([(1, 111)]))) == [111, 8, 9]
     @assert levels(setlevels!(copy(p), Dict([(1, 111), (8, NA)]))) == [111, 9]
+    # issue #201
+    @assert levels(setlevels!(@pdata([1.0, 2.0]), [3,4])) == [3.0, 4.0]
 
     y = @pdata [1, NA, -2, 1, NA, 4, NA]
     @assert isequal(unique(y), @pdata [1, NA, -2, 4])
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -399,6 +399,9 @@ function setlevels!{T,R}(x::PooledDataArray{T,R}, newpool::AbstractVector{T})
     end
 end
 
+setlevels!{T, R}(x::PooledDataArray{T, R},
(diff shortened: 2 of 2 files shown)


tlnagy:
@nalimilan @garborg If you all could merge this when you get a chance, that would be great. It might also help to tag a new release so that Gadfly's testing issues are resolved.

Tests are failing on nightly, but that looks unrated to this PR.

nalimilan:
Could you add a test for the new method too? Else it might break without anybody noticing.

tlnagy:
Good point, I added a test for change.

test/pooleddataarray.jl
```diff
@@ -27,6 +27,8 @@ module TestPDA
     @assert levels(setlevels!(copy(p), [1,8,9, 10])) == [1, 8, 9, 10]
     @assert levels(setlevels!(copy(p), Dict([(1, 111)]))) == [111, 8, 9]
     @assert levels(setlevels!(copy(p), Dict([(1, 111), (8, NA)]))) == [111, 9]
+    # issue #201
+    @assert levels(setlevels!(@pdata(ByteString["a", "b"]), ASCIIString["c","d"])) == ["c", "d"]
```
nalimilan: `ByteString` is going to go away soon. Could you use e.g. `Int` and `Float64` instead?

After that, I'll leave it open a bit to leave the opportunity for others to comment, and then I'll merge.

tlnagy:
Switched to using `Int` and `Float64` and squashed my changes.

PR #205, opened 2016-07-10
RFC: work around MethodErrors with Base.OneTo

tkelman:
trying to work around OneTo issue on nightly

### DIFF
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -78,8 +78,11 @@ function Base.to_index(A::DataArray)
 end
 
 
-if isdefined(Base, :checkindex)
+if isdefined(Base, :checkindex) && isdefined(Base, :AbstractUnitRange)
     Base.checkindex(::Type{Bool}, ::AbstractUnitRange, ::NAtype) =
+        throw(NAException("cannot index an array with a DataArray containing NA values"))
+elseif isdefined(Base, :checkindex)
+    Base.checkindex(::Type{Bool}, ::UnitRange, ::NAtype) =
         throw(NAException("cannot index an array with a DataArray containing NA values"))
 else
     Base.checkbounds(::Type{Bool}, sz::Int, I::AbstractDataVector{Bool}) = length(I) == sz
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -98,9 +98,9 @@ end
 import Base: index_shape, index_lengths, setindex_shape_check
 
 if isdefined(Base, :OneTo)
-    indexshape(x...) = Base.to_shape(index_shape(x...))
+    _index_shape(x...) = Base.to_shape(index_shape(x...))
 else
-    indexshape = index_shape
+    _index_shape = index_shape
 end
 
 # Fallbacks to avoid ambiguity
@@ -155,7 +155,7 @@ end
 end
 
 function _getindex{T}(A::DataArray{T}, I::@compat Tuple{Vararg{Union{Int,AbstractVector}}})
-    shape = indexshape(A, I...)
+    shape = _index_shape(A, I...)
     _getindex!(DataArray(Array(T, shape), falses(shape)), A, I...)
 end
 
--- a/src/indexing.jl
+++ b/src/indexing.jl
@@ -97,6 +97,12 @@ end
 
 import Base: index_shape, index_lengths, setindex_shape_check
 
+if isdefined(Base, :OneTo)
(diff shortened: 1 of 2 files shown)


nalimilan:
Thanks. Looks good to me, though I'd have called these `_broadcast_shape` and `_index_shape`.

tkelman:
Unfortunately it seems to break an `@inferred` test on release, and still has a bounds-checking issue on nightly.

src/indexing.jl
```diff
@@ -79,7 +79,7 @@ end
 
 
 if isdefined(Base, :checkindex)
-    Base.checkindex(::Type{Bool}, ::UnitRange, ::NAtype) =
+    Base.checkindex(::Type{Bool}, ::AbstractUnitRange, ::NAtype) =
```
tkelman: @timholy there will presumably be a window of base versions where checkindex is defined but AbstractUnitRange is not, so for the sake of bisecting base bugs I think this needs adjusting a little? I gave you collaborator access to my fork if that helps, I'm not a member of JuliaStats (are you?)

tkelman:
I'll make Milan's suggested renamings when I'm at a desk.

src/indexing.jl
ararslan: Looks like Tim isn't a JuliaStats member. I was recently made a member so I'd be happy to help out however I can.

tkelman: I was just gonna declare you my next-most-active surrogate for Milan while he's on vacation :sunny:

ararslan: Party on, Wayne

tkelman:
okay if this is green I think it's good to merge, and I suppose tag too since there haven't been many serious changes since the last tag

ararslan:
Yep, I agree it should be merged and tagged. Idk how permissions work since you aren't a JuliaStats member but if you can't tag it then I can tag it a little later (then go around frantically submitting and rebasing my PRs to bump the DataArrays requirement).

Thanks so much for doing this!

tkelman:
I could submit a tag PR to metadata, but someone with commit access would need to push the git tag

tkelman:
> submitting and rebasing my PRs to bump the DataArrays requirement

Which PR's? This fixes matters on 0.5, but I don't think there's anything in here as far as DataArrays API's that other packages would rely on?

ararslan:
I assume this would fix the issue that's breaking the DataFrames and GLM tests on 0.5. Regarding the PRs, my brain was stuck in yesterday when I had a couple of outstanding PRs. (Insert hackneyed joke about Mondays here)

tkelman:
Ah, I hope so. What you would do there is add upper bounds on the `julia` dependency of old package tags to reflect that they don't work with too-new versions of 0.5-dev, but people don't usually go to that level of detail on Julia version bounds.

tkelman:
annoying thing about squashes is they don't preserve multiple-author information very well

ararslan:
Oh you're right, I hadn't noticed that. That's too bad.

ararslan:
Is there anything that can be done that's somewhere between a straight merge and a squash-n-merge that would better preserve the author information? Or in cases like these should regular merges be preferred?

tkelman:
Depends what you care more about - skipping the intermediate commits (not all of which may have worked perfectly) in the future if you need to bisect through the history of the package, or preserving the authorship information. I don't think git allows you to have more than one author on a single commit, though you could manually edit the message to at least acknowledge multiple authors in the content of the log?

ararslan:
Should I revert the commit, modify the log to mention Tim, and recommit? It'd be nice to have him in there. In the future I'll make sure all contributors are mentioned in squashed commits before merging. Sorry for the oversight here.

tkelman:
I don't think he'll take it personally, and the more-linear history might help matters for some future bisect, you never know. You could also do the thing that Stefan has done on base a few times and make an empty merge commit that has both current master and the granular series of commits from this branch as parents. Though I don't know what that would do to bisects...

ararslan:
I'm not sure how to do that but I'd be happy to figure it out if that's what you think would be best.

tkelman:
I'm not positive how to do it either, I don't think it's worth worrying about.

ararslan:
Okay. Well, to anyone reading this, Tim Holy rocks and has made invaluable contributions to this PR! 🎸 :metal: 👏

timholy:
Thanks guys. As @tkelman surmises, credit is much less important than avoiding possibly-dangerous git shenanigans, so no sweat.

PR #206, opened 2016-07-18
Replace eltype_plus with promote_eltype_op

martinholters:
Needed due to pending JuliaLang/julia#17398. However, requires a version of Compat tagged that includes https://github.com/JuliaLang/Compat.jl/pull/252, tentatively assumed to be 0.8.6, hence WIP at the moment.

### DIFF
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -1,6 +1,7 @@
 using DataArrays, Base.@get!
 using Base.Broadcast: bitcache_chunks, bitcache_size, dumpbitcache,
-                      promote_eltype, broadcast_shape, eltype_plus
+                      promote_eltype, broadcast_shape
+using Compat: promote_eltype_op
 
 if isdefined(Base, :OneTo)
     _broadcast_shape(x...) = Base.to_shape(broadcast_shape(x...))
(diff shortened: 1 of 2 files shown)


src/broadcast.jl
```diff
@@ -295,7 +295,7 @@ end
 (.*)(A::(@compat Union{DataArray{Bool}, PooledDataArray{Bool}}), B::BitArray) = databroadcast(*, A, B)
 @da_broadcast_vararg (.*)(As...) = databroadcast(*, As...)
 @da_broadcast_binary (.%)(A, B) = databroadcast(%, A, B)
-@da_broadcast_vararg (.+)(As...) = broadcast!(+, DataArray(eltype_plus(As...), _broadcast_shape(As...)), As...)
+@da_broadcast_vararg (.+)(As...) = broadcast!(+, DataArray(Compat.promote_eltype_op(@functorize(+), As...), _broadcast_shape(As...)), As...)
```
ararslan: You could explicitly import `promote_eltype_op` from Compat to avoid having to qualify all of the calls. (Not necessary, just a suggestion.)

martinholters: I guess I'll change that and push when Compat has a new version tagged to re-trigger CI here once it makes sense.

martinholters:
Travis error looks unrelated at first glance. Something changed between 0.5.0-dev+5550 (version on AppVeyor) and  0.5.0-dev+5567 (Version on Travis) that might be causing this?

martinholters:
Confirmed locally: master gives the same test error on 0.5.0-dev+5567, so unrelated to this PR.

martinholters:
CI failure is due to #208.

tkelman:
tag when convenient?

ararslan:
Will do

PR #213, opened 2016-09-12
vcat of PooledDataVector might need to expand the reftype

gustafsson:
This feature could then be used to fix vcat of pooled columns in DataFrames as well, https://github.com/JuliaStats/DataFrames.jl/pull/990

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -831,10 +831,7 @@ function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
     pa = (p1, p2...)
     pool = unique(T[[p.pool for p in pa]...;])
 
-    idx = [begin
-        m = indexin(p.pool, pool)
-        m[p.refs]
-    end for p in pa]
+    idx = [indexin(p.pool, pool)[p.refs] for p in pa]
 
     refs = Array{DEFAULT_POOLED_REF_TYPE,N}([idx...;])
     PooledDataArray(RefArray(refs), pool)
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -143,9 +143,6 @@ for (f, basef) in ((:pdatazeros, :zeros), (:pdataones, :ones))
     end
 end
 
-# Pooled reference type
-reftype{T,R}(pa::PooledDataArray{T,R}) = R
-
 ##############################################################################
 ##
 ## Basic size properties of all Data* objects
@@ -831,18 +828,11 @@ function dropna{T}(pdv::PooledDataVector{T})
 end
 
 function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
-    for p in p2
-        @assert ndims(p) == N
-        @assert size(p)[2:end] == size(p1)[2:end]
-    end
+    pa = (p1, p2...)
+    pool = unique(T[[p.pool for p in pa]...;])
 
-    pa = PooledDataArray[p1, p2...]
-
-    pools = Vector{T}[p.pool for p in pa]
-    pool = levels(T[pools...;])
-
-    idx = [ begin
-        m = findat(pool, p.pool)
+    idx = [begin
+        m = indexin(p.pool, pool)
         m[p.refs]
     end for p in pa]
 
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -830,26 +830,23 @@ function dropna{T}(pdv::PooledDataVector{T})
     return res
 end
 
-function Base.vcat(pa::PooledDataArray...)
-    N = length(size(pa[1]))
-    for p in pa
-        @assert length(size(p))==N
+function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
+    dim{T2,R2,N2}(p::PooledDataArray{T2,R2,N2}) = N
+
+    for p in p2
+        @assert dim(p)==N
     end
 
-    pools = [p.pool for p in pa]
-    pool = levels([pools...;])
+    pa = PooledDataArray[p1, p2...]
 
-    # grow the reftype as much as needed
-    # unless one of the reftypes in 'pa' was big enough
-    ref_sz = [typemax(reftype(p)) for p in pa]
-    sz = maximum([length(pool), ref_sz...])
-    REFTYPE = compactreftype(sz)
+    pools = Vector{T}[p.pool for p in pa]
+    pool = levels(T[pools...;])
 
-    idx = map(pa) do p
+    idx = [ begin
         m = findat(pool, p.pool)
         m[p.refs]
-    end
+    end for p in pa]
 
-    refs = Array{REFTYPE,N}([idx...;])
+    refs = Array{DEFAULT_POOLED_REF_TYPE,N}([idx...;])
     PooledDataArray(RefArray(refs), pool)
 end
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -831,10 +831,8 @@ function dropna{T}(pdv::PooledDataVector{T})
 end
 
 function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
-    dim{T2,R2,N2}(p::PooledDataArray{T2,R2,N2}) = N
-
     for p in p2
-        @assert dim(p)==N
+        @assert ndims(p) == N
         @assert size(p)[2:end] == size(p1)[2:end]
     end
 
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -143,6 +143,9 @@ for (f, basef) in ((:pdatazeros, :zeros), (:pdataones, :ones))
     end
 end
 
+# Pooled reference type
+reftype{T,R}(pa::PooledDataArray{T,R}) = R
+
 ##############################################################################
 ##
 ## Basic size properties of all Data* objects
@@ -251,13 +254,15 @@ end
 ##
 ##############################################################################
 
-function compact{T,R<:Integer,N}(d::PooledDataArray{T,R,N})
-    sz = length(d.pool)
-
+function compactreftype(sz)
     REFTYPE = sz <= typemax(UInt8)  ? UInt8 :
               sz <= typemax(UInt16) ? UInt16 :
               sz <= typemax(UInt32) ? UInt32 :
                                       UInt64
+end
+
+function compact{T,R<:Integer,N}(d::PooledDataArray{T,R,N})
+    REFTYPE = compactreftype(length(d.pool))
 
     if REFTYPE == R
         return d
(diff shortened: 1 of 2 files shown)


nalimilan:
Unfortunately, I don't think we can choose the reftype dynamically, as it would introduce type instability. What we could do is use `widen` to return a reftype just as large as what could possibly be needed to store all levels, but not more.

Anyway, I wouldn't encourage you to spend too much time on DataArrays, are we're going to stop using them in DataFrames (https://github.com/JuliaStats/DataFrames.jl/pull/1008). You can have a look at CategoricalArrays.jl instead, were similar features are needed.

gustafsson:
`widen(UInt8)==UInt64`. So I think a hardcoded DEFAULT_POOLED_REF_TYPE makes more sense.

I also made some other changes to make the return type stable.

Thanks for reminding me that DataArrays is being deprecated.

src/pooleddataarray.jl
```diff
@@ -830,26 +830,23 @@ function dropna{T}(pdv::PooledDataVector{T})
...
-    # grow the reftype as much as needed
-    # unless one of the reftypes in 'pa' was big enough
-    ref_sz = [typemax(reftype(p)) for p in pa]
-    sz = maximum([length(pool), ref_sz...])
-    REFTYPE = compactreftype(sz)
+    pools = Vector{T}[p.pool for p in pa]
+    pool = levels(T[pools...;])
 
-    idx = map(pa) do p
+    idx = [ begin
```
nalimilan: Why change this?

nalimilan:
`widen(UInt8)` gives `UInt32` here, but yeah, we would have to design a custom function for it to return `UInt16` instead. The point is just that the type should only depend on types, not on dynamic properties like the number of levels.

src/pooleddataarray.jl
gustafsson: @code_warntype said it couldn't deduce the return type if idx was Any[] as returned by map.

gustafsson:
Right, we could write a special vcat but on the other hand I think it's fair to have to call compact again after vcat. What do you think?

nalimilan:
Given that the package is going to be deprecated soon, any correct solution is fine. But for CategoricalArrays, an optimization like this would be interesting.

nalimilan:
Are you sure this code works for arrays of any dimension? You can take inspiration from `typed_vcat` in Base to ensure it does. Also, tests should be included for matrices (at least).

(It would be cool if we could avoid rewriting `vcat` from scratch, but for now that doesn't seem possible.)

nalimilan:
Since you create copies of the input refs already, a simple and robust strategy would be to `convert` them to the wanted reftype, and then call `vcat` on them: most of the logic would be implemented in Base, and you wouldn't have to rewrite it.

gustafsson:
I added some tests, and thanks for mentioning `ndims`

test/pooleddataarray.jl
```diff
@@ -109,4 +109,22 @@ module TestPDA
...
     @test isequal(Base.ipermute!!(copy(pda), [6, 1, 3, 5, 2, 4]), @pdata(["A", "A", "B", "B", "C", NA]))
+
+    a1 = 1:200
+    a2 = 100:300
+    pa1 = PooledDataArray(a1);
+    pa2 = PooledDataArray(a2);
+    ca1 = compact(pa1);
+    ca2 = compact(pa2);
+    @test vcat(ca1, ca2) == vcat(a1, a2)
+    @test vcat(ca1, ca2) |> DataArrays.reftype == DataArrays.DEFAULT_POOLED_REF_TYPE
```
nalimilan: Instead of adding a `reftype` function just for this, better use `isa(vcat(ca1, ca2), DataArray{Int, DataArrays.DEFAULT_POOLED_REF_TYPE}}`. Also compute `vcat(ca1, ca2)` only once.

src/pooleddataarray.jl
```diff
@@ -829,3 +829,23 @@ function dropna{T}(pdv::PooledDataVector{T})
...
+function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
+    for p in p2
+        @assert ndims(p) == N
+        @assert size(p)[2:end] == size(p1)[2:end]
+    end
+
+    pa = PooledDataArray[p1, p2...]
+
+    pools = Vector{T}[p.pool for p in pa]
+    pool = levels(T[pools...;])
```
nalimilan: Why not just  `pool = unique(vcat([p.pool for p in pa]...))`?

src/pooleddataarray.jl
```diff
@@ -829,3 +829,23 @@ function dropna{T}(pdv::PooledDataVector{T})
...
     return res
 end
+
+function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
+    for p in p2
+        @assert ndims(p) == N
+        @assert size(p)[2:end] == size(p1)[2:end]
+    end
+
+    pa = PooledDataArray[p1, p2...]
```
nalimilan: A tuple should be more efficient and simpler: `pa = (p1, p2...)`.

src/pooleddataarray.jl
```diff
@@ -829,3 +829,23 @@ function dropna{T}(pdv::PooledDataVector{T})
...
+
+    pools = Vector{T}[p.pool for p in pa]
+    pool = levels(T[pools...;])
+
+    idx = [ begin
+        m = findat(pool, p.pool)
```
nalimilan: Use `indexin(p.pool, pool)` instead of `findat`, since the former is in Base.

test/pooleddataarray.jl
```diff
@@ -109,4 +109,22 @@ module TestPDA
...
+    @test vcat(ca1, pa2) |> DataArrays.reftype == DataArrays.DEFAULT_POOLED_REF_TYPE
+
+    a1 = zeros(2,3,4,5)
+    a2 = zeros(3,3,4,5)
+    a1[1:end] = 1:length(a1)
+    a2[1:end] = (1:length(a2)) + length(a1)
```
nalimilan: Would be good to keep common levels between the arrays to test that, e.g. by replacing `+ length(a1)` with `+ 10`. AFAICT this wouldn't have worked with the current code in the PR. Also, using `length(a2):1` instead of `1:length(a2)` would be a harder test than having levels ordered.

test/pooleddataarray.jl
```diff
@@ -109,4 +109,22 @@ module TestPDA
...
+    ca1 = PooledDataArray(a1) |> compact;
```
nalimilan: `compact(PooledDataArray(a1))` is a more common style.

test/pooleddataarray.jl
```diff
@@ -109,4 +109,22 @@ module TestPDA
...
+    ca2 = PooledDataArray(a2) |> compact;
+    @test vcat(ca1, ca2) == vcat(a1, a2)
```
nalimilan: Also test the result type using `isa`.

test/pooleddataarray.jl
```diff
@@ -109,4 +109,22 @@ module TestPDA
...
+    a2 = 100:300
+    pa1 = PooledDataArray(a1);
+    pa2 = PooledDataArray(a2);
+    ca1 = compact(pa1);
+    ca2 = compact(pa2);
+    @test vcat(ca1, ca2) == vcat(a1, a2)
+    @test vcat(ca1, ca2) |> DataArrays.reftype == DataArrays.DEFAULT_POOLED_REF_TYPE
+    @test vcat(ca1, pa2) |> DataArrays.reftype == DataArrays.DEFAULT_POOLED_REF_TYPE
+
+    a1 = zeros(2,3,4,5)
```
nalimilan: `Array(2,3,4,5)`, since you fill the array right below.

gustafsson:
Thanks for the feedback!

test/pooleddataarray.jl
gustafsson: Makes sense. It did actually work with +10 instead, but we didn't know that until that test was added. I agree that it was a poor test to only have unique ordered values.

src/pooleddataarray.jl
gustafsson: That's another gem in `Base` that I had missed!

src/pooleddataarray.jl
gustafsson: I'm curious, how is a tuple more efficient than a typed array?

src/pooleddataarray.jl
```diff
@@ -829,3 +829,21 @@ function dropna{T}(pdv::PooledDataVector{T})
     resize!(res, total)
     return res
 end
+
+function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
+    for p in p2
```
nalimilan: Sorry, forgot to ask again: is this check really needed, i.e. won't the `vcat` call below catch it? Anyway, `@assert` shouldn't be used for caller errors: if you keep these, throw an `ArgumentError` with a message instead.

src/pooleddataarray.jl
```diff
@@ -143,6 +143,9 @@ for (f, basef) in ((:pdatazeros, :zeros), (:pdataones, :ones))
     end
 end
 
+# Pooled reference type
+reftype{T,R}(pa::PooledDataArray{T,R}) = R
```
nalimilan: No longer needed.

src/pooleddataarray.jl
nalimilan: A tuple is immutable, so the compiler can get rid of it entirely, while the array needs to be allocated. Unlikely to be significant here, though.

test/pooleddataarray.jl
nalimilan: OK, I hadn't realized that the calls to `levels` was indeed equivalent to `unique`.

src/pooleddataarray.jl
gustafsson: I checked with code_warntype, and I had to keep some type hints to make the return type stable. But I also noticed that the code was shorter with `pa` as tuple than as an array.

src/pooleddataarray.jl
gustafsson: It will, but it would be harder for a user to figure out what was wrong in their arguments. Do you have a generic guideline to not validate arguments if it is guaranteed to crash anyways? :)

gustafsson: Actually the error you get from vcat of the refs below is really clear. I removed this check.

src/pooleddataarray.jl
nalimilan: AFAIK the type annotation in `PooledDataArray[p1, p2...]` doesn't improve type stability, since the type is abstract (lacks type parameters). On the contrary, it will make the array abstract even if type inference is table to identify a common concrete type for all elements.

Tuples don't have this problem as their type includes information regarding each element separately.

src/pooleddataarray.jl
```diff
@@ -829,3 +826,16 @@ function dropna{T}(pdv::PooledDataVector{T})
     resize!(res, total)
     return res
 end
+
+function Base.vcat{T,R,N}(p1::PooledDataArray{T,R,N}, p2::PooledDataArray...)
+    pa = (p1, p2...)
+    pool = unique(T[[p.pool for p in pa]...;])
+
+    idx = [begin
```
nalimilan: I would find it nicer as a single line `idx = [indexin(p.pool, pool)[p.refs] for p in pa]`.

nalimilan:
Thanks, looks good to me (apart from the minor style remark). I'll leave the PR open a bit more to let others comment.

BTW, I would really appreciate a PR adding this feature to CategoricalArrays. We really need the more efficient concatenation strategy implemented here.

src/pooleddataarray.jl
gustafsson: Thanks I learnt something new today!

The type annotation I was referring to that I wasn't able to get rid of was the `T` in this`pool = unique(T[[p.pool for p in pa]...;])`.

@code_warntype tells me that even if it knows the type of pa to be  `pa::Tuple{DataArrays.PooledDataArray{Int64,UInt8,1},DataArrays.PooledDataArray{Int64,UInt8,1}}` it will call regular vcat (instead of typed_vcat) and get a pool of type `pool::Union{Array{Any,1},Array{Int64,1}}`

PR #214, opened 2016-09-12
ErrorException doesn't take multiple arguments, but error does

gustafsson:
This error is never shown because it raises another error instead, namely that no matching constructor of ErrorException was found.

Also, int is deprecated in favour of Int.

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -487,10 +487,11 @@ function getpoolidx{T,R}(pda::PooledDataArray{T,R}, val::Any)
         push!(pda.pool, val)
         pool_idx = length(pda.pool)
         if pool_idx > typemax(R)
-            throw(ErrorException(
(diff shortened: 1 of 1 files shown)


simonster:
I think this should actually be `throw(ArgumentError(...))`.

gustafsson:
ArgumentError it is

```
julia> using DataArrays
julia> ca=compact(PooledDataArray(1:255));
julia> push!(ca,1423)
ERROR: ArgumentError: You're using a PooledDataArray with ref type UInt8,
which can only hold 255 values, and you just tried to add the 256th
reference. Please change the ref type to a larger int type, or use the
default ref type (UInt32).
 in getpoolidx at /Users/johan/.julia/v0.4/DataArrays/src/pooleddataarray.jl:490
 in push! at /Users/johan/.julia/v0.4/DataArrays/src/datavector.jl:138
```

ararslan:
Looks good.

PR #215, opened 2016-09-12
[rs] is deprecated, use [rs...] instead

gustafsson:


johnmyleswhite:
I suspect this should use `collect(rs)` to avoid a tuple-splitting penalty.

gustafsson:
@johnmyleswhite sure, that makes sense

@nalimilan did you change your opinion on `[a...]` vs `collect(a)`? https://github.com/JuliaStats/DataFrames.jl/pull/944

nalimilan:
> @nalimilan did you change your opinion on [a...] vs collect(a)? JuliaStats/DataFrames.jl#944

Not really. In the present case, splatting might involve thousands of values (think of `1:1000`), which will be very costly for the compiler. In the other issue, splatting only involved a few arguments which were already passed as varargs, so a method was compiled for each number of arguments anyway.

ararslan:
This looks good to me.

gustafsson:
@nalimilan I see, thanks for the explanation

nalimilan:
Thanks!

PR #217, opened 2016-09-16
fix anyna and allna

### LINKED ISSUE #216: `anyna` failure
Due to [this line](https://github.com/JuliaStats/DataArrays.jl/blob/master/src/abstractdataarray.jl#L88)
we have the following unexpected result
```julia
julia> anyna(Any[NA, 1])
false
```
We should probably define it as 
```julia
anyna(a::AbstractArray) =any(isna(a))
allna(a::AbstractArray) =all(isna(a))
```

panlanfeng:
Fix #216


ararslan:
It would be great if you could add tests for this.

nalimilan:
Thanks!

Issue #218, opened 2016-09-20
DataArrays MethodError on Julia 0.6 in DataFrames tests


ararslan:
The DataFrames tests currently fail on Julia nightly, but pass on 0.4 and 0.5, apparently due to an issue with DataArrays. See [here](https://travis-ci.org/JuliaStats/DataFrames.jl/jobs/161388884#L415) for a full Travis log. The relevant portion is reproduced below:

```
Error During Test
  Test threw an exception of type MethodError
  Expression: predict(m) == [ones(size(d,1)) d[:x1] d[:x2] d[:x1] .* d[:x2]] * collect(1:4)
  MethodError: no method matching broadcast_shape(::DataArrays.DataArray{Int64,1}, ::DataArrays.DataArray{Int64,1})
   in .*(::DataArrays.DataArray{Int64,1}, ::DataArrays.DataArray{Int64,1}) at /home/travis/.julia/v0.6/DataArrays/src/broadcast.jl:297
   in include_from_node1(::String) at ./loading.jl:541
   in macro expansion; at /home/travis/.julia/v0.6/DataFrames/test/runtests.jl:35 [inlined]
   in anonymous at ./<missing>:?
   in include_from_node1(::String) at ./loading.jl:541
   in process_options(::Base.JLOptions) at ./client.jl:262
   in _start() at ./client.jl:326
    FAILED: statsmodel.jl
LoadError: There was an error during testing
 in macro expansion; at /home/travis/.julia/v0.6/DataFrames/test/runtests.jl:43 [inlined]
 in anonymous at ./<missing>:?
 in include_from_node1(::String) at ./loading.jl:541
 in process_options(::Base.JLOptions) at ./client.jl:262
 in _start() at ./client.jl:326
while loading /home/travis/.julia/v0.6/DataFrames/test/statsmodel.jl, in expression starting on line 41
```

If this is a DataFrames problem rather than DataArrays, I can migrate this over there.

ararslan:
I don't think this is still applicable

PR #219, opened 2016-10-07
Replace calls to findat() with calls to indexin()

nalimilan:
The latter is provided in Julia Base whiel the former is deprecated in
StatsBase. Note that findat() returns the first matching index, but
indexin() the last one. Here it doesn't matter since pools only contains unique
values. Also remove an unnecessarily complex myunique() definition.

Fixes breakage (and warnings) introduced by incorrect deprecation in StatsBase. Cf. https://github.com/JuliaLang/METADATA.jl/pull/6660#issuecomment-252338632.


ararslan:
The nightly error looks like #218.

nalimilan:
Maybe just double-check with @johnmyleswhite that the `myunique` change is not missing something obvious. Anyway with https://github.com/JuliaStats/StatsBase.jl/pull/218 we can fix the breakage quickly, and this PR is only useful to get rid of deprecation warnings.

PR #222, opened 2016-11-03
Fix broadcast on master and some deprecations

andreasnoack:


nalimilan:
Thanks, these failures are really annoying for DataFrames. Though you'll need to check `VERSION` for `broadcast_indices` too. See https://github.com/JuliaStats/NullableArrays.jl/pull/164.

andreasnoack:
Done!

andreasnoack:
I'll do that.

test/reducedim.jl
```diff
@@ -127,7 +127,11 @@ for Areduc in (DataArray(rand(3, 4, 5, 6)),
             (1, 2, 3), (1, 3, 4), (2, 3, 4), (1, 2, 3, 4)]
             # println("region = $region, skipna = $skipna")
 
-            outputs = Any[DataArray(fill(NaN, Base.reduced_dims(size(Areduc), region)))]
+            if VERSION < v"0.6-"
```
tkelman: please be more specific about version cutoffs like this

REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 0.4
 StatsBase 0.3
-Compat 0.8.6
+Compat 0.9.3
```
tkelman: what needs 0.9.3? dot broadcast rewriting has been implemented for longer than that

test/reducedim.jl
andreasnoack: I missed this one. The other conditions in the PR are precise.

REQUIRE
andreasnoack: Nothing that I'm aware of but I thought I'd better bump it to match the version used for the testing.

Issue #226, opened 2016-12-30
Function isprime() not found in Julia v 0.6 1781


colbec:
`isprime()` is still found in Julia v 0.5 so DataArrays still works there.
In v 0.6 `isprime()` is not found unless we `Pkg.add("Primes")`, after which `isprime()` is found (isprime has been moved out of Base).
However precompile of DataArrays 0.3.10 fails with `ERROR: LoadError: LoadError: UndefVarError: isprime not defined`

tlnagy:
This is causing Gadfly's tests to fail on julia master.

ararslan:
Fixed by #227.

PR #227, opened 2016-12-30
Update predicates.jl for function isprime()

colbec:
isprime() has been moved to Primes from Base.
Ref: https://github.com/JuliaStats/DataArrays.jl/issues/226

### DIFF
--- a/src/predicates.jl
+++ b/src/predicates.jl
@@ -1,7 +1,7 @@
 if VERSION < v"0.6.0-dev+1781"
   numeric_predicates = (:(Base.iseven), :(Base.ispow2), :(Base.isfinite), :(Base.isprime), :(Base.isinf), :(Base.isodd))
 else
-  using Primes
+  import Primes
   numeric_predicates = (:(Base.iseven), :(Base.ispow2), :(Base.isfinite), :(Primes.isprime), :(Base.isinf), :(Base.isodd))
 end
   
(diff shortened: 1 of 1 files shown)


nalimilan:
Thanks, but you also need to add Primes.jl to `REQUIRE` and add a `import` call.

colbec:
Thanks, I needed the guidance. I have replaced my `using` statement with an `import`, and tried to update the `REQUIRE`.

ararslan:
I don't think we need Primes.jl here at all. It seems to me it would be sufficient to do:
```julia
numeric_predicates = [:(Base.iseven), :(Base.ispow2), :(Base.isfinite), :(Base.isinf), :(Base.isodd))]
isdefined(Base, :isprime) && push!(numeric_predicates, :(Base.isprime))
```

colbec:
I'm happy to yield to a better solution @ararslan . My suggestion was just a starting point.

ararslan:
If you'd be willing to use the code I posted in my comment above, I think that should do it. Thanks for your help here!

colbec:
I think the code offered by @ararslan is working, but there is a new error 

> no method matching size(::Type{DataArrays.DataArray})

In addition it would be really helpful to eliminate the large numbers of warnings related to broadcast functions.

nalimilan:
Honestly, I'm not sure somebody is going to do the work of fixing all the deprecations. We have already a lot of work to do to replace DataArrays with NullableArrays, so 0.5 might well be the last Julia release where the former will work without deprecation warnings. Fixing the error in `broadcast` would already be great.

PR #230, opened 2017-01-06
Fix reducedim tests for backports to 0.5

andreasnoack:
I think the backporting in https://github.com/JuliaLang/julia/pull/19103 broke the tests on 0.5.

### DIFF
--- a/test/reducedim.jl
+++ b/test/reducedim.jl
@@ -127,10 +127,10 @@ for Areduc in (DataArray(rand(3, 4, 5, 6)),
             (1, 2, 3), (1, 3, 4), (2, 3, 4), (1, 2, 3, 4)]
             # println("region = $region, skipna = $skipna")
 
-            if VERSION < v"0.6.0-dev.1121"
+            if VERSION < v"0.5.1-pre+19"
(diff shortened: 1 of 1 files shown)


andreasnoack:
Test failures are only on 0.6.

tkelman:
unreleased versions of branches have not been tested for stability and should not be coded against in packages

andreasnoack:
Makes sense. I'm using `release_05` branch for 0.5 so I didn't realize that this hadn't been released. Do you the part I coded against will change before the 0.5.1 release?

tkelman:
this isn't the only package where it caused issues, so likely

tkelman:
The place this should change is in Base's release-0.5, if there's any regression in functionality whatsoever between 0.5.0 and release-0.5, please open an issue and we'll fix it there. Making changes like this just makes it messier to catch real regressions when doing pkgeval comparisons when preparing patch releases.

tkelman:
backporting https://github.com/JuliaLang/julia/pull/19196 will make this unnecessary

andreasnoack:
Unnecessary or breaking?

tkelman:
hopefully just unnecessary but I'm not sure exactly what's being tested here

PR #235, opened 2017-02-12
Fixes for 0.6

andreasnoack:
The main parts are the fixes for the broadcasting and indexing which both were broken on Julia master. There are many fixes of deprecated syntax and I've changed some uses of the old `ngenerate` in favor of generated functions. I've decided to remove support for Julia version prior to 0.6 to make the source as simple as possible. The package has been undermaintained for a while so I think we should focus on making maintenance as easy as possible. Prior version fo Julia can just use old versions of this package.

@mbauman It would be appreciated if you could take a look at the indexing code. In particular, if it is really necessary with `@generated` in methods like https://github.com/JuliaStats/DataArrays.jl/blob/anj/06/src/indexing.jl#L164-L172 that used to be `@ngenerate`d.

@simonster You wrote most of the code touched here so any comments from you would also be appreciated.


andreasnoack:
Oh, and I can see that my build of Julia has become obsolete while finishing the PR.

src/DataArrays.jl
```diff
@@ -1,14 +1,14 @@
-__precompile__()
+__precompile__(false)
```
ararslan: What's keeping from being able to precompile?

src/broadcast.jl
```diff
@@ -84,271 +59,133 @@ end
...
-                    $(Symbol("na_$(k)")) = $(Symbol("A_$(k)")).na.chunks
-                    $(Symbol("data_$(k)")) = $(Symbol("A_$(k)")).data
-                    $(Symbol("state_$(k)_0")) = $(Symbol("state_$(k)_$(nd)")) = 1
-                    @nexprs $nd d->($(Symbol("skip_$(k)_d")) = size($(Symbol("data_$(k)")), d) == 1)
-                end : arrtype[k] == PooledDataArray ? quote
-                    $(Symbol("refs_$(k)")) = $(Symbol("A_$(k)")).refs
-                    $(Symbol("pool_$(k)")) = $(Symbol("A_$(k)")).pool
-                end : nothing
-            for k = 1:narrays]...))
+Base.map!{F}(f::F, B::Union{DataArray, PooledDataArray}, A0::AbstractArray, As::AbstractArray...) =
```
ararslan: Is the typevar necessary here?

simonster: yes, otherwise it won't specialize on the function

src/indexing.jl
```diff
@@ -187,25 +160,24 @@ function Base.getindex(pda::PooledDataArray, I::Real)
         return pda.pool[getindex(pda.refs, I)]
     end
 end
-@nsplat N function Base.getindex(pda::PooledDataArray, I::NTuple{N,Real}...)
-    if getindex(pda.refs, I...) == 0
-        return NA
-    else
-        return pda.pool[getindex(pda.refs, I...)]
+
+@generated function Base.getindex(pda::PooledDataArray, I::Integer...)
```
mbauman: I would just `@inline` this and the following.

test/operators.jl
```diff
@@ -2,6 +2,47 @@ module TestOperators
...
+                                  :(Base.log10),
+                                  :(Base.log1p),
+                                  :(Base.log2),
+                                  :(Base.exponent),
+                                  :(Base.sqrt),
+                                  :(Base.gamma),
+                                  :(Base.lgamma),
+                                  :(Base.digamma),
+                                  :(Base.erf),
+                                  :(Base.erfc)]
```
ararslan: No longer in Base (same with `erf` and `digamma`)

src/broadcast.jl
ararslan: Are you sure? I thought that was changed in Base a while back.

simonster: It's always possible I'm wrong :-D. But this is the situation described in https://github.com/JuliaLang/julia/issues/19137, and there's no indication the behavior has changed there.

ararslan: Ah, hmm. My bad then.

andreasnoack:
Thanks for the comments. I've updated the code accordingly. The latest changes on master haven't made it into the Julia binary yet so we'll have to restart the tests when new binaries have been uploaded.

src/indexing.jl
```diff
@@ -114,6 +114,8 @@ Base.getindex(t::AbstractDataArray, i::Real) =
 
 ## getindex: DataArray
 
+Base.linearindexing(x::Union{DataArray,PooledDataArray}) = Base.LinearFast()
```
mbauman: This should be defined on the `::Type{<:Union{DataArray,PooledDataArray}}`, IIRC.

ararslan:
All of the remaining depwarns are from other packages. Does anything else need to happen here before this can be merged? LGTM.

andreasnoack:
I think it is ready.

ararslan:
Awesome. Unless there are further comments, I say we should merge this soon.

src/DataArrays.jl
```diff
@@ -1,14 +1,14 @@
 __precompile__()
 
 module DataArrays
+    using Base: promote_op
     using Base.Cartesian, Compat, Reexport
```
ararslan: Since Compat was removed from REQUIRE, it should be removed here too. (Not sure how this didn't cause CI to fail...)

src/DataArrays.jl
```diff
@@ -1,14 +1,14 @@
...
-    import Compat.String
     @reexport using StatsBase
```
ararslan: This is tangential to the current changes, but it would make sense to me to stop reexporting StatsBase from this package. I'm not sure why we do it currently. If we do stop doing that, we could also drop the dependency on Reexport (not that it's a huge or problematic dependency, just saying).

src/DataArrays.jl
andreasnoack: True. I think it is installed by the dependencies.

ararslan: Ah yeah, probably installed by StatsBase.

src/broadcast.jl
nalimilan: AFAIK, without the `::F`, specialization will only happen if you call the function.

test/broadcast.jl
```diff
@@ -20,7 +20,7 @@ rb = 1:5
 @test broadcast!(+, DataArray(Int, 2, 2), [1, 0], [1  4]) == [2 5; 1 4]
 @test broadcast!(+, DataArray(Int, 2), [1, 0], [1, 4]) == [2, 4]
 @test broadcast!(+, DataArray(Int, 2), [1, 0], 2) == [3, 2]
-@test broadcast!(abs, @data([-1, -2])) == @data([1, 2])
+# @test broadcast!(abs, @data([-1, -2])) == @data([1, 2])
```
nalimilan: Why is this commented out? Should it be removed altogether?

andreasnoack: Yes. It should be removed. The in-place `broadcast!` is deprecated in Base.

andreasnoack:
I've added `Compat` back. We need it for the `@nsplat` and `@ngenerate` macros. Eventually, we should just use `@generated` but I'll prefer to make that change in a separate PR.

Tests are failing because of a problem with the Julia binaries. I suggest that we just wait and see if tomorrow's binaries are more up to date.

test/reducedim.jl
```diff
@@ -109,7 +109,7 @@ macro test_da_approx_eq(da1, da2)
         v2 = $(esc(da2))
         na = isna(v1)
         @test na == isna(v2)
-        defined = !na
+        defined = (!).(na)
```
ararslan: `.!na` should work here as you've written it elsewhere, right?

andreasnoack: Nope. Apparently, this form is required in macros. I tried the other version first.

ararslan: That's... bizarre. Maybe file a Base issue?

src/natype.jl
```diff
@@ -14,24 +14,28 @@
 ##
 ##############################################################################
 
-type NAtype
+struct NAtype
```
ararslan:
```
ERROR: LoadError: LoadError: syntax: extra token "NAtype" after end of expression
Stacktrace:
 [1] include_from_node1(::String) at ./loading.jl:539
 [2] include(::String) at ./sysimg.jl:14
 [3] include_from_node1(::String) at ./loading.jl:539
 [4] include(::String) at ./sysimg.jl:14
 [5] anonymous at ./<missing>:2
while loading /home/travis/.julia/v0.6/DataArrays/src/natype.jl, in expression starting on line 17
```
??????

andreasnoack: For some reason, the nightly build used an older commit than the previous one and therefore it doesn't include the `struct` changes. We can restart again when new binaries are up.

ararslan: I thought `type NAtype` was still supposed to work in 0.6? I guess not...?

ararslan: Looks like this is still happening with the nightly binaries. 😕

andreasnoack: See https://status.julialang.org/. They are old. There is a problem on master that prevents the buildbot from making new ones.

src/broadcast.jl
```diff
@@ -138,10 +140,16 @@ Base.map!{F}(f::F, B::Union{DataArray, PooledDataArray}, A0, As...) =
                     end : nothing
                 for k = 1:N]...))
 
-                # Extract values for ordinary AbstractArrays
+                # Extract values for other type
                 $(Expr(:block, [
-                    :(@inbounds $(Symbol("v_$(k)")) = @nref $nd $(Symbol("A_$(k)")) d->$(Symbol("j_$(k)_d")))
-                    for k = find(t -> !(t <: DataArray || t <: PooledDataArray), As)]...))
+                    As[k] <: AbstractArray  && !(As[k] <: Union{DataArray, PooledDataArray}) ? quote
```
ararslan: I think you could do `AbstractDataArray` instead of the `Union` there.

andreasnoack: Good point

andreasnoack:
I'll suggest that I test this one locally against latest Julia master and merge if the tests pass since we don't know when we'll have new nightlies for Travis testing.

ararslan:
Sounds like a good plan to me!

ararslan:
Thanks so much for your hard work here, Andreas!

nalimilan:
The use of `struct` makes the tests fail. Better use the old type declaration syntax for now.

ararslan:
It makes the tests fail because for some reason the nightly binaries are stuck at a commit before the Base change due to issues with the buildbots. Hopefully that should be fixed soon.

nalimilan:
OK. But I guess it won't fix the tests on Julia 0.5, will it?

ararslan:
Andreas bumped the minimum requirement for the package to 0.6-, so it won't work on 0.5 anyway.

Issue #92, opened 2014-05-19
Some methods from the standard library don't fulfil their contract


tomasaschan:
#### `unique(A::DataArray)` reorders items

According to [the spec](http://julia.readthedocs.org/en/latest/stdlib/base/#Base.unique), `unique(A)` should not re-order elements. However, it does for `DataArray`s:

```
julia> A = sort(rand(100)); # Get some data, sorted for simplicity
julia> DA = DataArray(A);
julia> all(diff(unique(A)) .>= 0) # Test that this works for regular arrays
true
julia> all(diff(unique(DA)) .>= 0) # It doesn't work for DataArrays!
false
julia> all(diff(unique(DA.data)) .>= 0) # It DOES work for the underlying data
true
```

Note that this is a problem even without any 'NA` values.

For sorted data this isn't really a big problem, as I can just re-sort it again, but for data that is not sorted but where ordering is important (such as experimental observations) that's not an option either.
#### `sort(A::DataArray)` throws error when missing values

Using `A` and `DA` as above, I get

```
julia> sort(DA)
100-element DataArray{Float64,1}:
 0.000314079
 0.00973866 
 # etc...
```

i.e. it Just Works(TM). However, if there are missing values:

```
julia> DA[35] = NA;
julia> sort(DA)
ERROR: no method isnan(ForwardOrdering, NAtype)
 in nans2right! at sort.jl:409
 in fpsort! at sort.jl:428
 in sort! at sort.jl:441
 in sort at sort.jl:334
```

I'd prefer if it "just worked" and placed all `NA` values last...

PR #237, opened 2017-02-19
Change unique() to return values in the same ordering as levels for PDAs

nalimilan:
While the generic unique() method says it preserves the order of appearance,
the ordering of levels is more likely to be useful. In particular, it will
allow StatsModels to use unique() to get levels present in the data in the
user-defined order, with the first level as reference by default.

The new code (inspired by CategoricalArrays) is also more efficient in the
common case where all values are encountered well before the end of the array,
by doing a periodic short-circuiting check.

-------
See https://github.com/JuliaStats/StatsModels.jl/pull/13#discussion_r101899349. The current behavior was chosen after discussion at  https://github.com/JuliaStats/DataArrays.jl/issues/92, but it looks like the issues where mainly about `DataArray`, not `PooledDataArray`. It's slightly annoying to deviate from the standard behavior of `unique`, but I can't think of cases where the order of appearance would be more useful than the ordering of levels, which should be carefully chosen (else using a PDA doesn't make much sense).

The other solution is to provide a separate function for this, but that sounds overkill.

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -271,9 +271,9 @@ end
 #' @description
 #'
 #' Return a DataVector containing the unique values of a `PooledDataArray`,
-#' in the order they appear in the data, including `NA` if any missing entries
+#' in the order of `levels`, including `NA` if any missing entries
 #' are encountered. For `PooledDataArray`s, this function is much less efficient
-#' than `levels`, which does not return the values in the same order.
+#' than `levels`.
 #'
 #' @param da::DataArray{T} `DataArray` whose unique values are desired.
 #'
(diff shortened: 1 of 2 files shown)


nalimilan:
Another approach which I find possibly better would be to keep the current behavior, but define `unique(x::AbstractArray, sort::Bool=false)`. When `sort=true`, the order of `levels` would be preserved for `PooledDataArray`; for other `AbstractArray`, we would simply call `sort!(unique(x))`. That would exactly suit our needs (see https://github.com/JuliaStats/StatsModels.jl/issues/14#issuecomment-281625381), and could be a useful behavior in general. The default method could even be defined in Base.

ararslan:
> The default method could even be defined in Base.

If we go that route I think it needs to be in Base, otherwise we're committing the grave sin of type piracy (i.e. overloading Base methods with Base types in an external package).

nalimilan:
Yet another approach is not to change anything, and obtain the ordered set of unique values via `intersect(levels(x), unique(x))`. The advantage is that it's simple and does not require a new API, the drawback is that `unique` does not return the most useful result by default (i.e. functions which call `unique` generically and return results based on that will use a meaningless order while a useful order is available).

nalimilan:
I think we should merge this and discuss the larger issue of what  `unique` means for `AbstractArray` later, as we need it to get StatsModels to work with DataArrays.

ararslan:
I'm hesitant to deviate from the behavior in Base when overloading a Base function, even if a more useful behavior is possible. `intersect(levels(x), unique(x))` seems okay to me, if a bit verbose.

nalimilan:
Yeah, but that requires defining `levels` somewhere. Not so urgent I guess if we go with  https://github.com/JuliaStats/DataFrames.jl/pull/1170 for now.

PR #238, opened 2017-03-07
RFC: The Great Doc-ening

ararslan:
This PR converts the many structured comments to actual docstrings as appropriate, with the overall goal of making full documentation for the package.

To-do list:
- [x] Finish documenting the functions
- [x] Set up Documenter
- [x] Glue together a cohesive manual

There are more files in this repo than I remembered.

### DIFF
--- a/docs/src/da.md
+++ b/docs/src/da.md
@@ -1,0 +1,40 @@
+# Representing missing data
+
+```@meta
+CurrentModule = DataArrays
+```
+
+```@docs
+NA
+NAtype
+```
+
+## Arrays with possibly missing data
+
+```@docs
+AbstractDataArray
+AbstractDataVector
+AbstractDataMatrix
+DataArray
+DataVector
+DataMatrix
+@data
+isna
+allna
+anyna
+dropna
+levels
+```
+
+## Pooled arrays
+
+```@docs
+PooledDataArray
+@pdata
+compact
+setlevels
+setlevels!
+replace!
+PooledDataVecs
+getpoolidx
+```
--- a/docs/src/da.md
+++ b/docs/src/da.md
@@ -23,6 +23,7 @@ isna
 allna
 anyna
 dropna
+padNA
 levels
 ```
 
@@ -37,4 +38,5 @@ setlevels!
 replace!
 PooledDataVecs
 getpoolidx
+reorder
 ```
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -1,8 +1,7 @@
 # TODO: Remove some T's from output type signatures
 
 """
-    DataArray{T,N}(d::Array{T,N}, m::BitArray{N} = falses(size(d)))
-    DataArray{T,N}(d::Array{T,N}, m::Array{Bool})
+    DataArray{T,N}(d::Array{T,N}, m::AbstractArray{Bool} = falses(size(d)))
 
 Construct a `DataArray`, an `N`-dimensional array with element type `T` that allows missing
 values. The resulting array uses the data in `d` with `m` as a bitmask to signify missingness.
@@ -59,7 +58,7 @@ function DataArray{T, N}(d::Array{T, N},
     return DataArray{T, N}(d, m)
 end
 
-function DataArray(d::Array, m::Array{Bool}) # -> DataArray{T}
+function DataArray(d::Array, m::AbstractArray{Bool}) # -> DataArray{T}
     return DataArray(d, BitArray(m))
 end
 
@@ -74,14 +73,14 @@ end
 """
     DataVector{T}
 
-A `DataArray` of order 1 with element type `T`.
+A 1-dimensional `DataArray` with element type `T`.
 """
 const DataVector{T} = DataArray{T, 1}
 
 """
     DataMatrix{T}
 
-A `DataArray` of order 2 with element type `T`.
+A 2-dimensional `DataArray` with element type `T`.
 """
 const DataMatrix{T} = DataArray{T, 2}
 
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -1,23 +1,35 @@
 # TODO: Remove some T's from output type signatures
 
-#' @description
-#'
-#' A type that extends the normal Julia Array{T} type by allowing
-#' the entries to be missing (i.e. NULL or NA) in addition to
-#' taking on values of type `T`.
-#'
-#' @field data::Array{T, N} The elements of the DataArray are stored here,
-#'        unless the given element is missing. In this case, the contents
-#'        of the corresponding entry in `data` can be any arbitrary value.
-#' @field na::BitArray{N} A 1-bit per entry Boolean mask specifying,
-#'        for each entry of the DataArray, whether that entry is `NA`
-#'        or not.
-#'
-#' @examples
-#'
-#' dv = DataArray([1, 2, 3], [false, false, true])
-#'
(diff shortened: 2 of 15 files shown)


src/natype.jl
```diff
@@ -14,9 +14,19 @@
...
+
+The data type of a missing value, `NA`.
+"""
 struct NAtype
 end
 
+"""
+    NA
+
+The sentinel value representing missingness.
```
nalimilan: I don't think "sentinel" applies here: it's used for values which are in the domain for a type, but which are used to denote missingness.

src/natype.jl
```diff
@@ -14,9 +14,19 @@
 ##
 ##############################################################################
 
+"""
+    NAtype
+
+The data type of a missing value, `NA`.
```
nalimilan: Not sure "data" really makes the description clearer.

src/dataarray.jl
```diff
@@ -42,104 +54,37 @@ mutable struct DataArray{T, N} <: AbstractDataArray{T, N}
...
-#' @returns out::DataArray A copy of `da`.
-#'
-#' @examples
-#'
-#' dv = @data [false, false, true, false]
-#' dv_new = copy(dv)
+"""
+    DataVector{T}
+
+A `DataArray` of order 1 with element type `T`.
```
nalimilan: "A one-dimensional `DataArray`" rather?

src/dataarray.jl
```diff
@@ -1,23 +1,35 @@
...
-#'        for each entry of the DataArray, whether that entry is `NA`
-#'        or not.
-#'
-#' @examples
-#'
-#' dv = DataArray([1, 2, 3], [false, false, true])
-#'
-#' dm = DataArray([1 2; 3 4], [false false; true false])
+"""
+    DataArray{T,N}(d::Array{T,N}, m::BitArray{N} = falses(size(d)))
```
nalimilan: Just `m::AbstractArray{Bool, N}` should be enough.

ararslan: Not entirely accurate though, since `DataArray{Bool,N}` is an `AbstractArray{Bool,N}`, but `m` can't be a `DataArray`.

src/natype.jl
ararslan: Yeah, agreed.

src/natype.jl
ararslan: Ah right. I'll update.

src/dataarray.jl
nalimilan: Right. Ah, well, we should probably widen the signature.

ararslan: Just allowing `Array{Bool}` and `BitArray` seems reasonable to me. What would be the benefit of allowing any array type here?

nalimilan: I think the burden of the proof should be rest on the other side: why no accept any `AbstractArray`, given that we convert the input to `BitArray` anyway?

ararslan: Fair point

src/datavector.jl
```diff
@@ -168,7 +168,24 @@ Base.sizehint!(pda::PooledDataVector, newsz::Integer) =
     sizehint!(pda.refs, newsz)
 
 # Pad a vector with NA's
-
+"""
+    padNA(dv::AbstractDataVector, front::Integer, back::Integer) -> DataVector
+
+Pad `dv` with `NA` values, `front` at the beginning of the array and `back` at the end.
```
nalimilan: Maybe make it more explicit that `front` and `back` are the number of elements.

src/pooleddataarray.jl
```diff
@@ -19,6 +19,52 @@ struct RefArray{R<:Integer,N}
...
+Construct a `PooledDataArray` based on the unique values in the given array. `PooledDataArray`s are
+useful for efficient storage of categorical data with a limited set of unique values.
+
+### Optional arguments
+
+* `pool`: The possible values of `data`. Defaults to `unique(data)`.
+* `m`: A missingness indicator akin to that of [`DataArray`](@ref). Defaults to `falses(size(d))`.
+* `r`: The integer subtype used to store pool references. Defaults to `$DEFAULT_POOLED_REF_TYPE`.
+
+---
```
nalimilan: I think the convention is just to skip a line such cases. You could also split the docstring into separate ones, one for each method.

src/pooleddataarray.jl
```diff
@@ -19,6 +19,52 @@ struct RefArray{R<:Integer,N}
...
+
+    PooledDataArray(T::Type, [R::Type=$DEFAULT_POOLED_REF_TYPE], [dims...])
+
+Construct a `PooledDataArray` with element type `T`, reference storage type `R`, and dimensions
+`dims`. If the dimensions are specified and nonzero, the array is filled with `NA` values.
+
+# Examples
+
+```jldoctest
```
nalimilan: I'd rather put this example below the description of the corresponding method.

src/pooleddataarray.jl
```diff
@@ -19,6 +19,52 @@ struct RefArray{R<:Integer,N}
...
+ "B"
+ "A"
+ "B"
+ "A"
+ "B"
+ "A"
+ "B"
+```
+
+Rather than storing 8 distinct strings, `p` stores 8 references to 2 distinct
```
nalimilan: This sentence is a bit out of place here.

src/pooleddataarray.jl
```diff
@@ -257,6 +303,46 @@ function compactreftype(sz)
...
+ "B"
+ "A"
+ "B"
+ "A"
+ "B"
+ "A"
+ "B"
+```
+
+Notice how the second type parameter is compacted from `$DEFAULT_POOLED_REF_TYPE` to
```
nalimilan: I think this kind of comment is usually put inline. Not sure.

src/pooleddataarray.jl
```diff
@@ -370,6 +419,12 @@ end
 myunique(x::AbstractVector) = unique(x)
 myunique(x::AbstractDataVector) = unique(dropna(x))
 
+"""
+    setlevels(x::PooledDataArray, newpool::Union{AbstractVector, Dict})
+
+Create a new [`PooledDataArray`](@ref) using the data in `x` with the value pool
```
nalimilan: Should mention (here and below) that this affects the element's values, not just the ordering of levels, i.e. the order of levels is what identifies them and allows replacing the old ones with new ones.

src/pooleddataarray.jl
```diff
@@ -481,6 +548,13 @@ Base.find(pdv::PooledDataVector{Bool}) = find(convert(Vector{Bool}, pdv, false))
 ##
 ##############################################################################
 
+"""
+    getpoolidx(pda::PooledDataArray, val)
+
+Return the index of the first occurrence of `val` in the value pool for `pda`.
```
nalimilan: I suspect "first occurrence of" is misleading, since I don't think duplicate values are allowed in pools.

docs/src/da.md
```diff
@@ -0,0 +1,42 @@
...
+## Pooled arrays
+
+```@docs
+PooledDataArray
+@pdata
+compact
+setlevels
+setlevels!
+replace!
+PooledDataVecs
```
ararslan: [Yes](https://github.com/JuliaStats/DataArrays.jl/blob/master/src/DataArrays.jl#L51). IMO this package exports way more than it should.

ararslan:
First round of comments addressed!

PR #240, opened 2017-03-11
Don't overload unsafe_getindex with Base types

ararslan:
Let's see if this breaks anything.


nalimilan:
This package really contains surprising things...

PR #242, opened 2017-03-19
Update tests to use testsets

ararslan:
This PR restructures the tests to use `@testset` and `@test` rather than wrapping units of tests in modules and checking correctness with `@assert`.

### DIFF
--- a/test/abstractarray.jl
+++ b/test/abstractarray.jl
@@ -1,14 +1,11 @@
-module TestAbstractArray
-    using Base.Test
-    using DataArrays
-
+@testset "AbstractArray" begin
     unsorted_dv = @data [2, 1, NA]
 
     # TODO: Make this work
     # tiedrank(dv)
 
-    @assert first(unsorted_dv) == 2
-    @assert isna(last(unsorted_dv))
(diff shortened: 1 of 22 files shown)


nalimilan:
Are the failures expected?

ararslan:
Expected? Not entirely. I may have done some bad find-and-replaces, or turned things into `@test`s when they shouldn't have been. Will look more at this in a bit.

ararslan:
Turns out the failures are macro things that don't like being in `@testset`s. Moving macro definitions outside and not using `@eval` in a macro in a macro seems to help.

ararslan:
Okay, failures are all fixed. The commits should be squashed on merge. I welcome anyone to take a look.

nalimilan:
I've only skimmed the changes, but this looks good to me as long as you only moved existing tests (I would have a hard time spotting subtle changes inside the diff!).

ararslan:
Yeah, the diff is a bit of a beast. Yes, I only moved existing tests; there should be no functional changes, just nicer test output. Thanks for taking a look!

PR #247, opened 2017-04-13
Add parent field to DataArray

quinnj:
Just like NullableArrays.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -32,8 +32,9 @@ julia> DataArray(Float64, 3, 3)
 mutable struct DataArray{T, N} <: AbstractDataArray{T, N}
     data::Array{T, N}
     na::BitArray{N}
+    parent::Vector{UInt8}
 
-    function DataArray{T,N}(d::Array{T, N}, m::BitArray{N}) where {T, N}
+    function DataArray{T,N}(d::Array{T, N}, m::BitArray{N}, parent::Vector{UInt8}=Vector{UInt8}()) where {T, N}
         # Ensure data values and missingness metadata match
         if size(d) != size(m)
             msg = "Data and missingness arrays must be the same size"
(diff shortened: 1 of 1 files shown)


quinnj:
@andreasnoack @nalimilan, any concerns here? This just mirrors what we have in NullableArrays in an effort to make DataArray-backed DataFrames 1st-class in the DataStreams world.

andreasnoack:
Could you explain why this is useful?

quinnj:
The general idea is that if you mmap some data, you have a `Vector{UInt8}` that you need to hold a reference to while your data in the actual `.values` field can reference that memory. The most immediate application is for `WeakRefString` which is just a `Ptr{UInt8}` and `len` into some data.

ararslan:
@simonster Can you explain your 👎?

simonster:
What happens if the DataArray gets garbage-collected but you still have an element from inside of it, and then you try to access that element?

quinnj:
Noodling on this today, I think a good solution might be to define `getindex(A::DataArray{WeakRefString}, i) => String`, i.e. if you pull an individual element out, it converts to a regular string. That would allow us to utilize the performance boosts of WeakRefStrings with more safety.

simonster:
That would be much better in terms of safety, but you'll be allocating every time you take a string out of the array, so the perf improvement may not be so impressive. Also you'd also want to make sure that anything that returns the data field of the DataArray (e.g. `convert(Array, x::DataArray)`) also does it safely.

In general, I think a better way of abstracting this may be to allow DataArrays to wrap arbitrary StridedArrays, and then implement `WeakRefArray{T,N} <: DenseArray{T,N}` and stick the parent field there.

nalimilan:
See previous discussion regarding wrapping arbitrary arrays here: https://github.com/JuliaStats/NullableArrays.jl/issues/112#issuecomment-225365429 What has changed is that we now have triangular dispatch in Julia 0.6.

andreasnoack:
> What has changed is that we now have triangular dispatch in Julia 0.6.

Even without that, it doesn't look like the problem here is any different from any of the other wrapper `AbstractArray` types we have had for a long time. It requires some care in the constructors but it seems to work well. I'd prefer allowing the values to be any `AbstractArray`.

PR #249, opened 2017-05-04
Remove a problematic test case that we do not want to support

ararslan:
Ref discussion in https://github.com/JuliaStats/StatsBase.jl/pull/248

### DIFF
--- a/test/reduce.jl
+++ b/test/reduce.jl
@@ -145,6 +145,5 @@ end
 
     da2[1:2:end] = NA
     keep = .!da1.na .& .!da2.na
-    @test isna(mean(da1, weights(da2)))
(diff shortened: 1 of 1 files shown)


ararslan:
> The fact that it was a single isolated line tends to confirm that this "feature" wasn't really considered as important.

Agreed. Especially because just below the behavior is shown to be correct when using the `skipna` argument, which is really what you'd want in that case anyway.

nalimilan:
I've just found out that the `WeightsVew` signature was relaxed on purpose in https://github.com/JuliaStats/StatsBase.jl/pull/89. It's still not clear to me what's the use case. @simonster?

PR #251, opened 2017-05-08
Fix isna(::DataArray) deprecation

nalimilan:
The deprecation wasn't used as the varargs `isa(::DataArray, ::Real...)` method
unexpectedly took precedence. It shouldn't apply when there are zero indices,
so change its signature to reflect this.

Fixes multiple test failures in dependencies discussed at https://github.com/JuliaStats/DataArrays.jl/pull/243, https://github.com/JuliaLang/METADATA.jl/issues/9148 and https://github.com/dmbates/MixedModels.jl/issues/79.

PR #252, opened 2017-05-08
Fix isna(::DataArray) deprecation (alternate take)

nalimilan:
The deprecation wasn't used as the varargs isa(::DataArray, ::Real...) method
unexpectedly took precedence. It shouldn't apply when there are zero indices,
so change its signature to reflect this.

Get rid of the @nsplat call which isn't needed here (and no longer used in
Julia Base).

-------------

This is a possibly cleaner alternative to https://github.com/JuliaStats/DataArrays.jl/pull/251. I don't understand why `@nsplat` was used here, but AFAICT we can remove it now since the function is so simple.

### DIFF
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -243,16 +243,13 @@ function Base.convert{T, N}(::Type{Array}, da::DataArray{T, N}, replacement::Any
 end
 
 dropna(dv::DataVector) = dv.data[.!dv.na] # -> Vector
-isna(da::DataArray, I::Real) = getindex(da.na, I)
 
 Base.broadcast(::typeof(isna), da::DataArray) = copy(da.na)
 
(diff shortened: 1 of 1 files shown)


andreasnoack:
This is much nicer but will there be a performance penalty for the splatting? I guess that was the reason why `@nsplat` was used in the old version.

nalimilan:
I have no idea,  but given that `getindex(::Array, ::Int, ::Int, ::Int...)` is defined in the same way I imagine this is no longer an issue.

ararslan:
Would it be possible to add a test that would have caught the issue?

nalimilan:
No, because the failure was in the deprecated method.

ararslan:
Ah right. Okay then. By bad for messing up the deprecation.

nalimilan:
Let's merge this one then? It can hardly be worse than the present state anyway. :-)

PR #254, opened 2017-05-24
Fix WeightVec deprecation from StatsBase 0.15

tkelman:


tkelman:
Could someone make a branch off of https://github.com/JuliaStats/DataArrays.jl/commit/2a83487bb7cbf570bb6679eb5995ce3545c2869c that supports Julia 0.5 so we could fix this there too?

ararslan:
That commit doesn't support 0.5, unless I'm missing something.

tkelman:
My mistake, copy-pasted the wrong thing. I meant the commit right before it, https://github.com/JuliaStats/DataArrays.jl/commit/2f235b4de970490a75acb970c55cd4660143367a

ararslan:
I assume the deprecation is causing issues on 0.5 in some packages that depend on DataArrays?

ararslan:
A `release-0.3` branch has been made for commits that support Julia 0.5, so we can release more 0.3.x versions from there.

ararslan:
The 32-bit AV commit looks unrelated, but I have absolutely no idea what could be causing it.

tkelman:
It's just introducing depwarn noise on 0.5 that wasn't there before. If the person who introduced that feature doesn't need it on 0.5, maybe it would be better to make that release of StatsBase 0.6-only.

One thing that's a bit annoying is StatsBase 0.15 doesn't support Julia 0.4 but DataArrays 0.3.x did. DataArrays 0.4.0 went straight to 0.6-only so there's no minor release series here that supports Julia 0.5 as a minimum. We could maybe tag it as DataArrays 0.4.1 which would be a bit out of order, or make a versioning exception and do 0.3.13 with a floor of Julia 0.5.

ararslan:
The Julia version jump between tags was out of necessity. Once we figured out how to get DataArrays working on 0.6 at all, we simply weren't able to find a way to support 0.5 and 0.6 at the same time, so we just had to drop everything but 0.6.

I don't know what the best way forward is in terms of backporting this. You know way more about that stuff than I do.

tkelman:
That's fine, but dropping support for 2 release series of Julia should have in retrospect bumped the package minor version twice, to leave a space for 0.5-only backports. If nobody cares much about the 0.4.0 release of DataArrays any more, it may be simplest to release the backport of this as 0.4.1.

ararslan:
So 0.4.0 will be support Julia 0.6 only, but 0.4.1 will support Julia 0.4 and 0.5?

tkelman:
No, DataArrays 0.4.1 would support Julia 0.5 and up (it can't support Julia 0.4 if it wants to depend on StatsBase 0.15, unless we go bring back 0.4 support there). Or just Julia 0.5, but it shouldn't make much difference since DataArrays 0.5.x is available for Julia 0.6.

ararslan:
Okay, sounds good. I've removed `release-0.3` and made a `release-0.4` branch.

tkelman:
Looks like the win32 issue is nightly-only https://ci.appveyor.com/project/nalimilan/dataarrays-jl/build/1.0.223 so that's mostly likely a bug in base julia. Worth attempting to isolate.

tkelman:
Should track the win32 issue separately, but I don't think this is causing it. The user experience of all these warnings is pretty bad, can we can these fixed?

ararslan:
Ah, sorry for the delay, been busy

PR #255, opened 2017-05-24
[release-0.4] backport of deprecation fix (#254) and CI updates (#253)

tkelman:


ararslan:
Hmmmmm those CI failures are concerning.

tkelman:
DataArrays has been failing tests on 0.5 on PkgEval all month - http://pkg.julialang.org/detail/DataArrays.html

Let's see if we can dig up what caused it.

tkelman:
https://github.com/JuliaLang/METADATA.jl/pull/9115

ararslan:
Ah! Can you add 36be3152e1ff785ce4015c0218a3a2c29c3e3465 to this as well then?

ararslan:
Oh, hm. That didn't seem to do it.

tkelman:
should track the failure on 0.5 separately, but this isn't making it any worse. can we get the deprecation fix released so `using DataFrames` doesn't warn for everyone?

PR #256, opened 2017-05-25
use 0.6.0-pre as minimum julia version

tkelman:
since struct syntax wouldn't work on early 0.6.0-dev versions,
better to stick with the julia-0.5-compatible versions of the package there

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.6-
+julia 0.6.0-pre
 Compat 0.8.6
 StatsBase 0.14.0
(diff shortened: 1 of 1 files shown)


ararslan:
This looks good, thanks. I figured we'd just bump it to 0.6 once that's released but this is probably a good idea in the interim.

What in the world is going on with our tests here? The failures are different across platforms...?!

ararslan:
cc @andreasnoack and @nalimilan: any ideas about what could be causing the failures?

PR #259, opened 2017-05-29
Add missing LineNumberNode to @swappable call (fixes #257)

### LINKED ISSUE #257: Precompile fails on latest 0.7 - type Symbol has no field head
Something happened in the last 24 hours. DataArrays was fine yesterday for precompile but now fails with `type Symbol has no field head` on line 580 of operators.jl which shows `@eval begin`.

```
> versioninfo()
Julia Version 0.7.0-DEV.357
Commit 9e3318c984 (2017-05-25 19:37 UTC)
Platform Info:
  OS: Linux (x86_64-suse-linux)
  CPU: Intel(R) Core(TM) i5-4460  CPU @ 3.20GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, haswell)

> using CUDAdrv, CUDAnative, DataArrays, DataFrames, PyCall, PyPlot,  Images, ImageView
INFO: Precompiling module DataArrays.
ERROR: LoadError: LoadError: type Symbol has no field head
Stacktrace:
 [1] macro expansion at /home/colin/.julia/v0.7/DataArrays/src/operators.jl:580 [inlined]
 [2] anonymous at ./<missing>:?
 [3] include_from_node1(::String) at ./loading.jl:551
 [4] include(::String) at ./sysimg.jl:14
 [5] include_from_node1(::String) at ./loading.jl:551
 [6] include(::String) at ./sysimg.jl:14
 [7] anonymous at ./<missing>:2
while loading /home/colin/.julia/v0.7/DataArrays/src/operators.jl, in expression starting on line 579
while loading /home/colin/.julia/v0.7/DataArrays/src/DataArrays.jl, in expression starting on line 66
ERROR: Failed to precompile DataArrays to /home/colin/.julia/lib/v0.7/DataArrays.ji.
Stacktrace:
 [1] compilecache(::String) at ./loading.jl:685
 [2] _require(::Symbol) at ./loading.jl:490
 [3] require(::Symbol) at ./loading.jl:398

```

getzdan:
There has been some change to Julia in the macrocall AST, adding the LineNumberNode.
Unless there is another change to Julia, some packages using `:macrocall` expression generation will break.

### DIFF
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -138,7 +138,11 @@ macro dataarray_binary_scalar(vectorfunc, scalarfunc, outtype, swappable)
                 if swappable
                     # For /, Array/Number is valid but not Number/Array
                     # All other operators should be swappable
-                    map!(x->Expr(:macrocall, Symbol("@swappable"), LineNumberNode(@__LINE__), x, scalarfunc), fns, fns)
+                    if VERSION < v"0.7.0-"
+                        map!(x->Expr(:macrocall, Symbol("@swappable"), x, scalarfunc), fns, fns)
+                    else
+                        map!(x->Expr(:macrocall, Symbol("@swappable"), LineNumberNode(@__LINE__), x, scalarfunc), fns, fns)
+                    end
                 end
                 Expr(:block, fns...)
             end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -138,7 +138,7 @@ macro dataarray_binary_scalar(vectorfunc, scalarfunc, outtype, swappable)
                 if swappable
                     # For /, Array/Number is valid but not Number/Array
                     # All other operators should be swappable
-                    map!(x->Expr(:macrocall, Symbol("@swappable"), x, scalarfunc), fns, fns)
+                    map!(x->Expr(:macrocall, Symbol("@swappable"), LineNumberNode(@__LINE__), x, scalarfunc), fns, fns)
                 end
                 Expr(:block, fns...)
             end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -138,7 +138,7 @@ macro dataarray_binary_scalar(vectorfunc, scalarfunc, outtype, swappable)
                 if swappable
                     # For /, Array/Number is valid but not Number/Array
                     # All other operators should be swappable
-                    if VERSION < v"0.7.0-"
+                    if VERSION < v"0.7.0-DEV.328"
                         map!(x->Expr(:macrocall, Symbol("@swappable"), x, scalarfunc), fns, fns)
                     else
                         map!(x->Expr(:macrocall, Symbol("@swappable"), LineNumberNode(@__LINE__), x, scalarfunc), fns, fns)


nalimilan:
Thanks. Indeed this fixes loading the package, but I wonder whether the test failure may be related: the path between `#=` and `=#` is suspicious.
```
Booleans: Error During Test
  Expression evaluated to non-Boolean
  Expression: any(#= C:\Users\appveyor\.julia\v0.7\DataArrays\test\booleans.jl:15 =# @data([NA, 1, 2]) .== 1) == true
       Value: NA
```

getzdan:
Don't think the test error is related. Actually, it seems to me the test should Fail (but not error), as 1 == NA should be NA. But this discussion is for the other issue.

src/operators.jl
```diff
@@ -138,7 +138,11 @@ macro dataarray_binary_scalar(vectorfunc, scalarfunc, outtype, swappable)
                 if swappable
                     # For /, Array/Number is valid but not Number/Array
                     # All other operators should be swappable
-                    map!(x->Expr(:macrocall, Symbol("@swappable"), LineNumberNode(@__LINE__), x, scalarfunc), fns, fns)
+                    if VERSION < v"0.7.0-"
```
nalimilan: It would be great if you could identify the exact commit which broke this (use `contrib/commit-name.sh`).

getzdan:
You are right `.==` is acting weird:
```
julia> (@data [NA, 1, 2]) .== 1
0-dimensional DataArrays.DataArray{Bool,0}:
NA
```

src/operators.jl
getzdan: My guess is the exact commit is: 0.7.0-DEV.328 . Specifically, https://github.com/JuliaLang/julia/commit/fcdf4376415fde3b76faeb1c6f6841735497c8ea

nalimilan:
I guess we can still merge this (after updating the version check) and continue debugging, as it's always easier to debug a package that can be loaded.

ararslan:
Thanks for figuring this out!

nalimilan:
The failure actually comes from a change in Base, see https://github.com/JuliaLang/julia/pull/22019 and https://github.com/JuliaLang/julia/pull/22130.

src/operators.jl
tkelman: You should use the merge commit https://github.com/JuliaLang/julia/commit/9e3318c9840e7a9e387582ba861408cefe5a4f75. Its parent commit, https://github.com/JuliaLang/julia/commit/8b2cac4215976c656c5a3337edd4f36eb1c8aa51, is 0.7.0-DEV.353 but does not have this change.

nalimilan: Damn, everybody gets this wrong at least once. Do you think the script could do this automatically?

tkelman: we could maybe have it do some git operations to try to figure out whether the commit it runs on is a merge (or squash merge) commit or something from a PR branch and give a warning, but I'm not sure what to check for

tkelman: If github would fix "rebase and merge" to include a reference to the PR number, we could switch to that for multi-commit PR's and disable the normal merge commit option. Single-commit PR's should almost always be squash merged since there's no reason not to, so it's only multi-commit PR's where there's a risk of VERSION getting things wrong.

tkelman: did this get fixed?

nalimilan: No, but see https://github.com/JuliaStats/DataArrays.jl/pull/264.

PR #266, opened 2017-06-23
Fix uninitialised values in unique for PooledDataArray

### LINKED ISSUE #265: #undef/uninitialised values from unique on PooledDataArray
Julia 0.6 and DataArrays v0.5.3 (also observed on Julia 0.5 and DataArrays v0.4.1)

If all elements in a level for a `PooledDataArray` are `NA`, then `unique` returns `#undef` for `String` elements and uninitialised values for `Int` elements. If only some of the elements for a level are `NA` (i.e. not all) then the issue is not observed.

### String
```julia
using DataArrays
x = PooledDataArray(["A", "B", "A"]);
masks = [[1], [2], [3], [1, 3]]
for mask in masks
    y = x[:]; y[mask] = NA
    @show mask
    display(unique(y))
    println()
end
```
gives:
```
mask = [1]
3-element DataArrays.DataArray{String,1}:
 NA
 "B"
 "A"

mask = [2]
3-element DataArrays.DataArray{String,1}:
    "A"
 #undef
 #undef

mask = [3]
3-element DataArrays.DataArray{String,1}:
 "A"
 "B"
 NA

mask = [1,3]
3-element DataArrays.DataArray{String,1}:
    NA
    "B"
 #undef
```

### Int
```julia
x = PooledDataArray([1, 2, 1]);
masks = [[1], [2], [3], [1, 3]]
for mask in masks
    y = x[:]; y[mask] = NA
    @show mask
    display(unique(y))
    println()
end
```
gives:
```julia
mask = [1]
3-element DataArrays.DataArray{Int64,1}:
  NA
 2
 1

mask = [2]
3-element DataArrays.DataArray{Int64,1}:
          1
 4492641712
 4491970448

mask = [3]
3-element DataArrays.DataArray{Int64,1}:
 1
 2
  NA

mask = [1, 3]
3-element DataArrays.DataArray{Int64,1}:
           NA
          2
 4491970448
```

### Version Info
```
Julia Version 0.6.0
Commit 903644385b (2017-06-19 13:05 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin13.4.0)
  CPU: Intel(R) Core(TM)2 Duo CPU     P7550  @ 2.26GHz
  WORD_SIZE: 64
  BLAS: libopenblas (USE64BITINT DYNAMIC_ARCH NO_AFFINITY Penryn)
  LAPACK: libopenblas64_
  LIBM: libopenlibm
  LLVM: libLLVM-3.9.1 (ORCJIT, penryn)
```

joshbode:
Use the number of observed values (plus one for `NA`) to determine the
length of the result of `unique` on `PooledDataArrays`.

Fixes #265

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -334,7 +334,8 @@ function Base.unique{T}(pda::PooledDataArray{T})
     end
 
     if firstna > 0
-        res = DataArray(Vector{T}(nlevels + 1))
+        n = length(unique_values)
+        res = DataArray(Vector{T}(n + 1))
         i = 0
         for val in unique_values
             i += 1
@@ -345,8 +346,8 @@ function Base.unique{T}(pda::PooledDataArray{T})
             res.data[i] = val
         end
 
-        if firstna == nlevels + 1
-            res.na[nlevels + 1] = true
+        if firstna == n + 1
+            res.na[n + 1] = true
         end
 
         return res
--- a/test/pooleddataarray.jl
+++ b/test/pooleddataarray.jl
@@ -126,4 +126,18 @@
     r = vcat(ca1, ca2)
     @test r == vcat(a1, a2)
     @test isa(r, PooledDataArray{Int,DataArrays.DEFAULT_POOLED_REF_TYPE,4})
+
+    # Issue #265:
+    #   Ensure that any levels that have are not expressed in a
+    #   PooledDataArray are handled correctly and not left uninitialised when
+    #   using unique()
+    x = ["A", "B", "A"];
+    masks = [[1], [2], [3], [1, 3]]
+    for mask in masks
+        y = PooledDataArray(x)
+        y[mask] = NA
+        @test isequal(sort(unique(y)), sort(DataArray(unique(y))))
+    end
+    z = PooledDataArray([1, 2], [1, 2, 3])
+    @test sort(unique(z)) == DataArray([1, 2])
 end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -6,8 +6,6 @@ const unary_vector_operators = [:(Base.median),
 
 # TODO: dist, iqr
 
-const ffts = [:(Base.fft)]
-
 const binary_vector_operators = [:(Base.dot),
                                  :(Base.cor),
                                  :(Base.cov),
@@ -21,7 +19,6 @@ const rowwise_operators = [:rowminimums,
                            :rowmedians,
                            :rowstds,
                            :rowvars,
-                           :rowffts,
                            :rownorms]
 
 # Swap arguments to fname() anywhere in AST. Returns the number of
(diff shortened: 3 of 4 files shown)


test/runtests.jl
```diff
@@ -3,6 +3,9 @@
 #
 
 using Base.Test
+if VERSION >= v"0.7.0-"
+    using FFTW
```
joshbode: Added FFTW for `0.7.0-` since it has been removed from Base, however the nightly build image doesn't support it yet

nalimilan: I would do `if !isdefined(Base, :FFTW)`, or check the exact commit when it was removed.

test/pooleddataarray.jl
```diff
@@ -126,4 +126,13 @@
     r = vcat(ca1, ca2)
     @test r == vcat(a1, a2)
     @test isa(r, PooledDataArray{Int,DataArrays.DEFAULT_POOLED_REF_TYPE,4})
+
+    # Issue #265
```
nalimilan: Would be nice to describe the issue in a few words.

src/pooleddataarray.jl
```diff
@@ -334,7 +334,8 @@ function Base.unique{T}(pda::PooledDataArray{T})
     end
 
     if firstna > 0
-        res = DataArray(Vector{T}(nlevels + 1))
+        n = length(unique_values)
```
nalimilan: Could define `n` earlier, since `length(unique_values)` is used above.

test/runtests.jl
nalimilan: Also, can you put this in a separate commit?

nalimilan:
Thanks, looks good apart from the two minor points I highlighted.

src/pooleddataarray.jl
joshbode: Isn't `length(unique_values)` potentially still varying above?

test/runtests.jl
joshbode: Will do. Any thoughts on what should be done in [src/operators.jl](https://github.com/JuliaStats/DataArrays.jl/blob/master/src/operators.jl#L9)?

nalimilan: Why not use the same check?

joshbode: What if FFTW.jl isn't already imported?
(unless we make it a dependency, which I don't think is a good idea)

nalimilan: I see. Let's ask @ararslan, but I would be inclined to deprecate this feature. People can convert to a standard array just like they need to do for any other external function.

ararslan: Yeah, let's get rid of anything FFT-related in this package.

joshbode: OK - my second commit should have got it all

ararslan: Fantastic, thanks!

PR #267, opened 2017-07-03
Fix bitwise operator ambiguities, test for ambiguities, deprecate vectorized ops

ararslan:
This should fix the ambiguity error that occurs when attempting to `&`, `|`, or `xor` two `DataArray`s. This also deprecates that functionality in favor of dot `broadcast`. To top it off, I've added ambiguity detection to the tests.

cc @nalimilan, I think this is what you were running into in DataFramesMeta.

### DIFF
--- a/src/deprecated.jl
+++ b/src/deprecated.jl
@@ -58,3 +58,19 @@ function xtabs{T}(x::AbstractArray{T})
     end
     return d
 end
+
+# Implicitly vectorized bitwise operations
+
+import Base: &, |, xor
+
+for f in [:(&), :(|), :(xor)]
+    for T in [:(BitArray), :(Range{<:Integer}), :(AbstractArray{<:Integer}), :(Integer)]
+        @eval begin
+            @deprecate ($f)(a::DataArray{<:Integer}, b::$T) ($f).(a, b)
+            @deprecate ($f)(a::$T, b::DataArray{<:Integer}) ($f).(a, b)
+        end
+    end
+    @eval begin
+        @deprecate ($f)(a::DataArray{<:Integer}, b::DataArray{<:Integer}) ($f).(a, b)
+    end
+end
--- a/src/operators.jl
+++ b/src/operators.jl
@@ -372,24 +372,6 @@ for f in (:(&), :(|), :(Base.xor))
     end
 end
 
-# # DataArray with DataArray
-# (&)(a::DataArray{Bool}, b::DataArray{Bool}) =
-#     DataArray(a.data & b.data, (a.na & b.na) | (a.na & b.data) | (b.na & a.data))
-# (|)(a::DataArray{Bool}, b::DataArray{Bool}) =
-#     DataArray(a.data | b.data, (a.na & b.na) | (a.na & !b.data) | (b.na & !a.data))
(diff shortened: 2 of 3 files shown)


src/operators.jl
```diff
@@ -369,24 +369,15 @@ for f in (:(&), :(|), :(Base.xor))
...
-#     DataArray(a.data $ b.data, a.na | b.na)
 
-# DataArray with non-DataArray
-# Need explicit definition for BitArray to avoid ambiguity
-for t in (:(BitArray), :(Range{Bool}), :(Union{AbstractArray{Bool}, Bool}))
-    @eval begin
-        @swappable (&)(a::DataArray{Bool}, b::$t) = DataArray(convert(Array{Bool}, a.data & b), a.na & b)
-        @swappable (|)(a::DataArray{Bool}, b::$t) = DataArray(convert(Array{Bool}, a.data | b), a.na & !b)
-        @swappable ($)(a::DataArray{Bool}, b::$t) = DataArray(convert(Array{Bool}, a.data $ b), copy(a.na))
+        function Base.broadcast(::typeof($f), a::DataArray{T}, b::S) where {T<:Integer, S<:Integer}
```
andreasnoack: Why can't they use the normal `broadcast` method?

ararslan: Because I was overthinking this 🙃

ararslan:
Okay, cool. The test failures now are because we have an enormous amount of ambiguities.

andreasnoack:
> The test failures now are because we have an enormous amount of ambiguities.

Please make sure that they are worth fixing before spending hours and memory on fixing, i.e. can the ambiguities actually happen with reasonable user programs?

ararslan:
Tbh I was just going to mark the test as broken rather than trying to fix all the ambiguities... I just wanted to make sure that the deprecation was actually taking effect rather than being lost in an ambiguity error, and that does appear to be the case now. I assume that if someone runs into an ambiguous case, they'll tell us about it.

src/deprecated.jl
```diff
@@ -58,3 +58,16 @@ function xtabs{T}(x::AbstractArray{T})
...
+
+# Implicitly vectorized bitwise operations
+
+import Base: &, |, xor
+
+for T in [:(DataArray{<:Integer}), :(BitArray), :(Range{<:Integer}), :(AbstractArray{<:Integer}), :(Integer)]
+    for f in [:(&), :(|), :(xor)]
+        @eval begin
+            @deprecate ($f)(a::DataArray{<:Integer}, b::$T) ($f).(a, b)
+            @deprecate ($f)(a::$T, b::DataArray{<:Integer}) ($f).(a, b)
```
ararslan: Now the only problem is that one of these definitions is causing method overwritten warnings, but it's hard to tell what's actually going on because it's saying that something on line 56 is overwriting something on line 56, and that line number is clearly wrong.

ararslan:
Cool, this should be good to go now.

nalimilan:
Cool, that's definitely nicer to help transition than the ambiguity error. Worth a new release?

ararslan:
Yeah I'd say so.

ararslan:
Let's wait for #270 before tagging

PR #268, opened 2017-07-04
Minor 0.6 cleanup

ararslan:
This change ~~removes the dependency on Compat (which was actually unused),~~ updates the type definition syntax to `struct`, and bumps the required Julia version from a 0.6 prerelease to 0.6 release.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,4 +1,4 @@
-julia 0.6.0-pre
+julia 0.6
 Compat 0.8.6
 StatsBase 0.15.0
 Reexport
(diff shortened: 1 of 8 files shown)


ararslan:
Oh, booooooo `@ngenerate` is from Compat, so we can't drop the dependency

andreasnoack:
We should get rid of `@ngenerate`

ararslan:
I agree but I don't understand it well enough to know how to replace it.

ararslan:
Okay, I've left the Compat dependency alone for now, so the change here is now quite minor.

PR #270, opened 2017-07-04
Use generated functions instead of @ngenerate macro

### LINKED ISSUE #269: Get rid of @ngenerate
This is a relic from Julia versions past, available only through Compat. We can ditch the Compat dependency altogether if we can remove the use of `@ngenerate`, which is currently only used for `mapreducedim`.

andreasnoack:
Fixes #269

### DIFF
--- a/src/reducedim.jl
+++ b/src/reducedim.jl
@@ -5,23 +5,25 @@ using Base.check_reducedims
 # This is a substantially faster implementation of the "all" reduction
 # across dimensions for reducing a BitArray to an Array{Bool}. We use
 # this below for implementing MaxFun and MinFun with skipna=true.
-@ngenerate N typeof(R) function Base._mapreducedim!{N}(f, op::typeof(&), R::Array{Bool}, A::BitArray{N})
-    lsiz = check_reducedims(R, A)
-    isempty(A) && return R
-    @nextract N sizeR d->size(R, d)
-    @nexprs 1 d->(state_0 = state_{N} = 1)
-    @nexprs N d->(skip_d = sizeR_d == 1)
-    Achunks = A.chunks
-    k = 1
-    @nloops(N, i, A,
-        d->(state_{d-1} = state_d),
-        d->(skip_d || (state_d = state_0)),
-        begin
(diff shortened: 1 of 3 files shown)


src/reducedim.jl
```diff
@@ -5,23 +5,25 @@ using Base.check_reducedims
...
-    @nloops(N, i, A,
-        d->(state_{d-1} = state_d),
-        d->(skip_d || (state_d = state_0)),
-        begin
-            @inbounds R[state_0] &= f(Base.unsafe_bitgetindex(Achunks, k))
-            state_0 += 1
-            k += 1
-        end)
-    R
+@generated function Base._mapreducedim!{N}(f, op::typeof(&), R::Array{Bool}, A::BitArray{N})
```
ararslan: Can the `{N}` be removed in favor of a trailing `where N`?

ararslan:
Well that's a _much_ simpler change than I was imagining. Thanks so much!

ararslan:
Want to drop the Compat dependency while you're in here?

src/reducedim.jl
```diff
@@ -5,23 +5,25 @@ using Base.check_reducedims
...
-    @nloops(N, i, A,
-        d->(state_{d-1} = state_d),
-        d->(skip_d || (state_d = state_0)),
-        begin
-            @inbounds R[state_0] &= f(Base.unsafe_bitgetindex(Achunks, k))
-            state_0 += 1
-            k += 1
-        end)
-    R
+@generated function Base._mapreducedim!(f, op::typeof(&), R::Array{Bool}, A::BitArray{N}) where {N}
```
tkelman: why is this here and not base? this is type piracy

andreasnoack: I guess this definition is older than the concept of type piracy. @simonster why didn't you submit this one to base?

tkelman: ping, is it still necessary to have this here?

ararslan: No, it should just be able to be removed here, though it would be good to know whether moving the code to Base would provide a performance improvement.

tkelman: then let's delete this

PR #272, opened 2017-07-05
Use `⊻` instead of deprecated `$`

omus:


src/operators.jl
```diff
@@ -359,12 +359,12 @@ end
...
 @swappable (&)(a::NAtype, b::Bool) = b ? NA : false
 @swappable (|)(a::NAtype, b::Bool) = b ? true : NA
-@swappable ($)(a::NAtype, b::Bool) = NA
+@swappable (⊻)(a::NAtype, b::Bool) = NA
 
 # To avoid ambiguity warning
 @swappable (|)(a::NAtype, b::Function) = NA
 
-for f in (:(&), :(|), :(Base.xor))
+for f in (:(&), :(|), :(⊻))
```
ararslan: This is equivalent to `Base.xor`

omus: Yes it is. I just updated it for consistency with the change above.

ararslan:
Thanks, much appreciated!

Issue #273, opened 2017-07-05
MethodError in pooleddataarray.jl in Julia 0.6


LouisMagarshack:
Hi all,

I'm seeing a method error when trying to use `groupby` and `unstack` on a DataFrame with an `AbstractString` column.

eg:

```julia
x = DataFrame(a = AbstractString["a"], b = [1.])
groupby(x, :a)
```

results in

```
MethodError: no method matching DataArrays.PooledDataArray(::Array{AbstractString,1}, ::Array{String,1}, ::BitArray{1}, ::Type{UInt32})
Closest candidates are:
  DataArrays.PooledDataArray(::AbstractArray{T,N}, ::Array{T,1}, ::AbstractArray{Bool,N}, ::Type{R<:Integer}) where {T, R<:Integer, N} at /home/ubuntu/.julia/v0.6/DataArrays/src/pooleddataarray.jl:113
  DataArrays.PooledDataArray(::AbstractArray{T,N}, ::Array{T,1}, ::AbstractArray{Bool,N}) where {T, N} at /home/ubuntu/.julia/v0.6/DataArrays/src/pooleddataarray.jl:113
  DataArrays.PooledDataArray(::DataArrays.RefArray{R<:Integer,N}, ::Array{T,1}) where {T, R<:Integer, N} at /home/ubuntu/.julia/v0.6/DataArrays/src/pooleddataarray.jl:102
  ...

Stacktrace:
 [1] DataArrays.PooledDataArray(::Array{AbstractString,1}, ::BitArray{1}, ::Type{UInt32}) at /home/ubuntu/.julia/v0.6/DataArrays/src/pooleddataarray.jl:144
 [2] DataArrays.PooledDataArray(::DataArrays.DataArray{AbstractString,1}) at /home/ubuntu/.julia/v0.6/DataArrays/src/pooleddataarray.jl:164
 [3] groupby(::DataFrames.DataFrame, ::Array{Symbol,1}) at /home/ubuntu/.julia/v0.6/DataFrames/src/groupeddataframe/grouping.jl:92
 [4] groupby(::DataFrames.DataFrame, ::Symbol) at /home/ubuntu/.julia/v0.6/DataFrames/src/groupeddataframe/grouping.jl:115
```

However 
```julia
x = DataFrame(a = String["a"], b = [1.])
groupby(x, :a)
```
works fine.

It also seems to be fixed by changing line 113 of pooleddataarray.jl from
```julia
[...]

PR #274, opened 2017-07-06
Fix MethodError when calling groupby, unstack, etc

LouisMagarshack:
See https://github.com/JuliaStats/DataArrays.jl/issues/273 .

### DIFF
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -137,7 +137,7 @@ end
 function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
                                          m::AbstractArray{Bool, N},
                                          r::Type{R} = DEFAULT_POOLED_REF_TYPE)
-    pool = convert(Array, unique(d[.!m]))
+    pool = convert(typeof(d), unique(d[.!m]))
     if method_exists(isless, (T, T))
         sort!(pool)
     end
--- a/src/pooleddataarray.jl
+++ b/src/pooleddataarray.jl
@@ -137,7 +137,7 @@ end
 function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
                                          m::AbstractArray{Bool, N},
                                          r::Type{R} = DEFAULT_POOLED_REF_TYPE)
-    pool = convert(Vector{eltype(d)}, unique(d[.!m]))
+    pool = convert(Vector{T}, unique(d[.!m]))
     if method_exists(isless, (T, T))
         sort!(pool)
     end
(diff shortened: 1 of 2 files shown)


ararslan:
Fantastic, thank you! Could you add a test that would catch this?

LouisMagarshack:
I'll add some tests and check why Travis is unhappy.

src/pooleddataarray.jl
```diff
@@ -137,7 +137,7 @@ end
 function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
                                          m::AbstractArray{Bool, N},
                                          r::Type{R} = DEFAULT_POOLED_REF_TYPE)
-    pool = convert(Array, unique(d[.!m]))
+    pool = convert(typeof(d), unique(d[.!m]))
```
ararslan: Ah, it actually looks like this isn't quite correct, since there are too many things that can be `AbstractArray` but for which `convert`ing like this doesn't make sense. That's why a test is failing; it's trying to `convert` to a `UnitRange`. How about
```julia
pool = convert(Array{eltype(d), ndims(d)}, unique(d[.!m]))
```
?

ararslan: Actually `ndims` shouldn't be preserved in this case, since `unique` always returns a `Vector`. So perhaps
```julia
pool = convert(Array{eltype(d)}, unique(d[.!m]))
```

LouisMagarshack: Yep you're right, I just ran into that `UnitRange` test issue!

LouisMagarshack:
My pleasure : D !

src/pooleddataarray.jl
```diff
@@ -137,7 +137,7 @@ end
 function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
                                          m::AbstractArray{Bool, N},
                                          r::Type{R} = DEFAULT_POOLED_REF_TYPE)
-    pool = convert(Array, unique(d[.!m]))
+    pool = convert(Vector{eltype(d)}, unique(d[.!m]))
```
andreasnoack: You can use `T` here instead of `eltype(d)`.

LouisMagarshack: Good idea, should be done 👌

src/pooleddataarray.jl
```diff
@@ -137,7 +137,7 @@ end
 function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
                                          m::AbstractArray{Bool, N},
                                          r::Type{R} = DEFAULT_POOLED_REF_TYPE)
-    pool = convert(Array, unique(d[.!m]))
+    pool = convert(Vector{T}, unique(d[.!m]))
```
nalimilan: How did this end up being `Vector{T}` rather than `Array{T}`? The comments seemed to recommend the latter. AFAICT that won't work for multidimensional arrays.

ararslan: `unique` always returns a `Vector`, even when the input is multidimensional, so this should be safe.

PR #277, opened 2017-07-18
Use the default Base implementation of map

### LINKED ISSUE #276: Inference failure/indexing issues
Maybe a DataFrames issue?

```julia
julia> d = DataFrame(a = [1, 2]);

julia> d[d[:a] .> 1, :]
1×1 DataFrames.DataFrame
│ Row │ a │
├─────┼───┤
│ 1   │ 2 │

julia> d[map(a -> a > 1, d[:a]), :]
ERROR: MethodError: no method matching getindex(::DataFrames.DataFrame, ::DataArrays.DataArray{Any,1}, ::Colon)
```

ararslan:
This ensures that we get the correct element type, among other things. Fixes #276.

### DIFF
--- a/src/datavector.jl
+++ b/src/datavector.jl
@@ -100,16 +100,6 @@ function Base.deleteat!(dv::DataVector, inds)
     dv
 end
 
-# TODO: should this be an AbstractDataVector, so it works with PDV's?
-function Base.map(f::Function, dv::DataVector)
-    n = length(dv)
-    res = DataArray(Any, n)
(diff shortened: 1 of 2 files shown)


andreasnoack:
I think we should leave these open for a bit longer

ararslan:
ok

tkelman:
ref https://github.com/JuliaLang/METADATA.jl/pull/10328 is there any objection to this being released?

PR #280, opened 2017-07-29
RFC: Stop lying about eltype

andreasnoack:
As discussed on Slack. It was actually not that bad. The core changes are in `abstractdataarray.jl` and defines
```julia
abstract type AbstractDataArray{T, N} <: AbstractArray{Union{T,NAtype}, N} end

Base.eltype{T, N}(d::AbstractDataArray{T, N}) = Union{T,NAtype}
```
So for subtypes of `AbstractDataArray`, the `T` is not referring to union but for signatures using `AbstractArray` the `T` refers to the union. This requires a little care when we ping-ping between Base and DataArrays definitions such as `broadcast` and `similar`. For that, it was convenient to define an `extractT` function to return the `T` part of a `Union{T,NAtype}`.

The changes here also require some changes to `DataFrames`. Hopefully, I'll be able to open a PR with them tomorrow and then we can look at the whole picture.

cc: @iamed2, @jeffbezanson

Update: Realized that an `eltype` definition is not needed here anymore since the fallback for `AbstractArray`s would now be correct.

### DIFF
--- a/src/abstractdataarray.jl
+++ b/src/abstractdataarray.jl
@@ -4,7 +4,7 @@
 An `N`-dimensional `AbstractArray` whose entries can take on values of type
 `T` or the value `NA`.
 """
-abstract type AbstractDataArray{T, N} <: AbstractArray{T, N} end
+abstract type AbstractDataArray{T, N} <: AbstractArray{Union{T, NAtype}, N} end
 
 """
     AbstractDataVector{T}
@@ -20,7 +20,7 @@ A 2-dimensional [`AbstractDataArray`](@ref) with element type `T`.
 """
 const AbstractDataMatrix{T} = AbstractDataArray{T, 2}
 
-Base.eltype(d::AbstractDataArray{T, N}) where {T, N} = T
+Base.eltype(d::AbstractDataArray{T, N}) where {T, N} = Union{T,NAtype}
 
 # Generic iteration over AbstractDataArray's
 
--- a/src/abstractdataarray.jl
+++ b/src/abstractdataarray.jl
@@ -4,7 +4,7 @@
 An `N`-dimensional `AbstractArray` whose entries can take on values of type
 `T` or the value `NA`.
 """
-abstract type AbstractDataArray{T, N} <: AbstractArray{Union{T, NAtype}, N} end
+abstract type AbstractDataArray{T, N} <: AbstractArray{Data{T}, N} end
 
 """
     AbstractDataVector{T}
--- a/src/broadcast.jl
+++ b/src/broadcast.jl
@@ -189,9 +189,30 @@ Base.Broadcast._containertype(::Type{T}) where T<:DataArray           = DataArra
 Base.Broadcast._containertype(::Type{T}) where T<:PooledDataArray     = PooledDataArray
 Base.Broadcast.broadcast_indices(::Type{T}, A) where T<:AbstractDataArray = indices(A)
 
+@inline function broadcast_t(f, ::Type{T}, shape, A, Bs...) where {T}
+    dest = Base.Broadcast.containertype(A, Bs...)(extractT(T), Base.index_lengths(shape...))
+    return broadcast!(f, dest, A, Bs...)
+end
+
+# This is mainly to handle isna.(x) since isna is probably the only
+# function that can guarantee that NAs will never propagate
+@inline function broadcast_t(f, ::Type{Bool}, shape, A, Bs...)
+    dest = similar(BitArray, shape)
+    return broadcast!(f, dest, A, Bs...)
+end
+
+# This one is almost identical to the version in Base and can hopefully be
+# removed at some point. The main issue in Base is that it tests for
+# isleaftype(T) which is false for Union{T,NAtype}. If the test in Base
+# can be modified to cover simple unions of leaftypes then this method
+# can probably be deleted and the two _t methods adjusted to match the Base
+# invokation from Base.Broadcast.broadcast_c
 @inline function Base.Broadcast.broadcast_c{S<:AbstractDataArray}(f, ::Type{S}, A, Bs...)
     T     = Base.Broadcast._broadcast_eltype(f, A, Bs...)
     shape = Base.Broadcast.broadcast_indices(A, Bs...)
-    dest = S(T, Base.index_lengths(shape...))
-    return broadcast!(f, dest, A, Bs...)
+    return broadcast_t(f, T, shape, A, Bs...)
 end
+
+# This one is much faster than normal broadcasting but the method won't get called
+# in fusing operations like (!).(isna.(x))
+Base.broadcast(::typeof(isna), da::DataArray) = copy(da.na)
--- a/src/dataarray.jl
+++ b/src/dataarray.jl
@@ -162,7 +162,7 @@ function Base.resize!(da::DataArray{T,1}, n::Int) where T
 end
 
 function Base.similar(da::DataArray, T::Type, dims::Dims) #-> DataArray{T}
-    return DataArray(Array{T}(dims), trues(dims))
+    return DataArray(Array{extractT(T)}(dims), trues(dims))
 end
 
 Base.size(d::DataArray) = size(d.data) # -> (Int...)
@@ -243,8 +243,6 @@ function Base.convert{T, N}(::Type{Array}, da::DataArray{T, N}, replacement::Any
 end
 
 dropna(dv::DataVector) = dv.data[.!dv.na] # -> Vector
-
-Base.broadcast(::typeof(isna), da::DataArray) = copy(da.na)
 
 Base.any(::typeof(isna), da::DataArray) = any(da.na) # -> Bool
 Base.all(::typeof(isna), da::DataArray) = all(da.na) # -> Bool
--- a/src/natype.jl
+++ b/src/natype.jl
@@ -36,11 +36,32 @@ struct NAException <: Exception
 end
 NAException() = NAException("NA found")
 
+# Restrict to Number to avoid infinite recursion
+## Numbers
+Base.promote_rule(::Type{Union{T,NAtype}}, ::Type{Union{S,NAtype}}) where {T<:Number,S<:Number} =
+    Union{promote_type(T, S),NAtype}
+Base.promote_rule(::Type{Union{T,NAtype}}, ::Type{S}) where {T<:Number,S<:Number} =
+    Union{promote_type(T, S),NAtype}
+## Dates
+Base.promote_rule(::Type{Union{T,NAtype}}, ::Type{Union{S,NAtype}}) where {T<:Dates.AbstractTime,S<:Dates.AbstractTime} =
+    Union{promote_type(T, S),NAtype}
+Base.promote_rule(::Type{Union{T,NAtype}}, ::Type{S}) where {T<:Dates.AbstractTime,S<:Dates.AbstractTime} =
+    Union{promote_type(T, S),NAtype}
+
+# Restrict to Number to avoid maching everything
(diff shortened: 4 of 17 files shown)


src/dataarray.jl
```diff
@@ -244,8 +244,6 @@ end
 
 dropna(dv::DataVector) = dv.data[.!dv.na] # -> Vector
 
-Base.broadcast(::typeof(isna), da::DataArray) = copy(da.na)
```
nalimilan: Isn't it worth keeping this specialization for performance? Remember that we have deprecated `isna(da)` in favor of `isna.(da)`.

src/natype.jl
```diff
@@ -36,12 +36,26 @@ struct NAException <: Exception
 end
 NAException() = NAException("NA found")
 
+Base.promote_rule(::Type{Union{T,NAtype}}, ::Type{Union{S,NAtype}}) where {T<:Number,S<:Number} =
```
nalimilan: I suppose these are only to fix ambiguities? If that's the case, better add a comment. If not, then why not support more general promotions? Same remark for `convert`.

src/pooleddataarray.jl
```diff
@@ -106,9 +106,11 @@ end
 PooledDataArray(d::PooledDataArray) = d
 
 # Constructor from array, w/ pool, missingness, and ref type
-function PooledDataArray{T,R<:Integer,N}(d::AbstractArray{T, N},
+function PooledDataArray{T,R<:Integer,N}(d::Union{AbstractArray{T, N},
+                                                  AbstractArray{Union{T,NAtype}, N}},
                                          pool::Vector{T},
-                                         m::AbstractArray{Bool, N},
+                                         m::Union{AbstractArray{Bool, N},
```
nalimilan: Could be written more simply as `<:Union{Bool, NAtype}`. You could also adopt the same approach for `d` and use `T = extractT(eltype(d))` inside the function body.

src/statistics.jl
```diff
@@ -39,25 +39,25 @@ gl(n::Integer, k::Integer) = gl(n, k, n*k)
...
 function StatsBase.describe{T<:Real}(io::IO, X::AbstractDataVector{T})
-    nacount = sum(isna.(X))
+    nacount = sum(isna, X)
     pna = 100nacount/length(X)
     if pna != 100 # describe will fail if dropna returns an empty vector
         describe(io, dropna(X))
     else
         println(io, "Summary Stats:")
-        println(io, "Type:           $(eltype(X))")
+        println(io, "Type:           $(T)")
```
nalimilan: Why not use `extractT(eltype(X))` just like below?

test/nas.jl
```diff
@@ -62,4 +62,11 @@
     @test_throws NAException for v in each_failna(dv); end
     @test collect(each_dropna(dv)) == a
     @test collect(each_replacena(dv, 4)) == [4, 4, a..., 4]
+
+    @testset "promotion" begin
+        @test promote_type(Int, Union{Float64,NAtype}) == Union{Float64,NAtype}
+        @test promote_type(Union{Int,NAtype}, Float64) == Union{Float64,NAtype}
+        @test promote_type(Union{Int,NAtype}, Union{Float64,NAtype}) == Union{Float64,NAtype}
```
nalimilan: Should also tests with non-number types, e.g. `String`.

test/broadcast.jl
```diff
@@ -128,4 +128,9 @@
     @test map!(abs, x, x) == @data([1, 2])
     @test isequal(map!(+, DataArray(Float64, 3), @data([1, NA, 3]), @data([NA, 2, 3])), @data([NA, NA, 6]))
     @test map!(isequal, DataArray(Float64, 3), @data([1, NA, NA]), @data([1, NA, 3])) == @data([true, true, false])
+
+    # isna doesn't propagate NAs so it should return BitArrays
+    x = (!).(isna.(@data [NA, 1, 2]))
```
nalimilan: Maybe also test without `(!).`? The code paths used are quite different due to fusion.

src/broadcast.jl
```diff
@@ -189,9 +189,26 @@ Base.Broadcast._containertype(::Type{T}) where T<:DataArray           = DataArra
...
 Base.Broadcast.broadcast_indices(::Type{T}, A) where T<:AbstractDataArray = indices(A)
 
+@inline function broadcast_t(f, ::Type{T}, shape, A, Bs...) where {T}
+    dest = Base.Broadcast.containertype(A, Bs...)(extractT(T), Base.index_lengths(shape...))
+    return broadcast!(f, dest, A, Bs...)
+end
+
+# This is mainly to handle isna.(x) since isna is probably the only
+# function that can guarantee that NAs will never propagate
+@inline function broadcast_t(f, ::Type{Bool}, shape, A, Bs...)
```
andreasnoack: No because the Base method is not extended. I just use the same name. See the comment above `broadcast_c` for an explanation. Base's broadcast needs some adjustments before we can use it here.

src/dataarray.jl
andreasnoack: I'd prefer to clean up the code. I'm not sure how big the performance difference is but even if there is a slight difference, I'm not sure if it really matters here. Having these kinds of definitions is misleading and I wasted some time because of it since `isna.(x)` and `(!).(isna.(x))` were returning different types.

src/natype.jl
andreasnoack: Yes. The restriction to `Number` was necessary. Otherwise, it kept calling itself. I can add a comment.

andreasnoack: The `convert` method is just to define convert in that case, not so fix an ambiguity.

src/statistics.jl
andreasnoack: The type parameter is already in the signature so I think it is easier and clearer just to use it.

test/nas.jl
andreasnoack: I'm not sure if there are any promotion methods for `String`.

src/dataarray.jl
nalimilan: Then at least measure the performance difference? `isna` is quite commonly used.

test/nas.jl
nalimilan: But why define methods for Number and not for other types? That was my main interrogation.

nalimilan: Ah, I guess you meant that it does not really make sense to promote `String` with anything else? Then you could try with e.g. `Dates.Minute` and `Dates.Second`.

src/statistics.jl
nalimilan: OK, I had misread `AbstractDataArray` as `AbstractArray`.

src/natype.jl
nalimilan: That explains why you wrote it that way, but doesn't justify why we'd want to support only `Number`. We should find a solution for other types. Nulls.jl [only defines](https://github.com/JuliaData/Nulls.jl/blob/master/src/Nulls.jl#L48-L49) methods for `Null` (equivalent of `NAtype`): couldn't we do the same here?

src/dataarray.jl
andreasnoack: Fair call. It is indeed much faster. I'll add it back but in the `broadcast.jl` file instead of here.

test/nas.jl
andreasnoack: I wanted to define if for any type but `Union{Any,NAtype}==Any` so that doesn't work. Basically, I don't know how we can cover all relevant cases except by adding them one by one. `Number` is probably the most category to cover and the only one covered by tests but I think you are right that `Dates` should also have a definition.

nalimilan: See my comment above about how Nulls.jl does it. We really need a fully generic solution.

andreasnoack: I don't think that is the same thing. Just tried it out to be sure. There will be call like these and the `Null`s definition won't help here. It will still return `Any`.

andreasnoack:
Thanks for the comments. I'm not sure how it will feel like in user code. We'll probably have to try it out a bit. I made the changes to `DataFrames` to get a feel for how much would have to be changed in library code. I don't think it was that bad and it will hopefully open up for using `DataArrays` in function not written explicitly for them.

My understanding is that `Null` is basically a reimplementation of `NAtype` with the main benefit that it lives in a separate package instead of being bundled with two array types. I can see that the semantics for comparisons are a bit different from `NA`. I could try making a commit with that change but I'm not sure it is worth bundling the two changes into the same PR.

ararslan:
> I could try making a commit with that change but I'm not sure it is worth bundling the two changes into the same PR.

👍 I think we should go ahead with this and try out the Nulls change in a separate branch. Luckily since the name is different (`null` vs `NA`), we should be able to put some deprecations in place that will ensure people make the right comparisons for `null`.

test/nas.jl
nalimilan: Hm, indeed.

@quinnj Had you noticed that `promote_type(Union{Int,Null}, Union{Float64,Null}) === Any` with Nulls.jl? Any ideas how to make it return `Union{Float64,Null}` in a fully generic way which would could e.g. with time types?

nalimilan:
Sure, I didn't mean we should move to Nulls.jl in this PR, just that it could make sense to make both changes before tagging a release (if we decide to do both).

test/nas.jl
quinnj: I hadn't noticed, but probably because I use [my own custom promoting machinery](https://github.com/JuliaData/CSV.jl/blob/jq/gangy/src/io.jl#L177) in CSV.jl (the one place I rely on promoting, as far as I know). I need my own rules, because I want things like `promote_type(::Type{Int}, ::Type{Date}) = String`. 

I did add the definition `promote_type2{T, S}(::Type{Union{T, Null}}, ::Type{Union{S, Null}}) = Union{promote_type2(T, S), Null}`; would that be enough to solve the case here?

andreasnoack: You last definition is what I tried first (with `pomote_rule`) but I got an infinite recursion out of it.

nalimilan: It seems to work here. Can you post exactly what you tried?

andreasnoack: Ah, so
```julia
promote_rule(::Type{Union{T,NAtype}}, ::Type{Union{S,NAtype}}) where {T<:Dates.AbstractTime,S<:Dates.AbstractTime} = Union{promote_type(T, S),NAtype}
```
seems fine but
```julia
promote_rule(::Type{Union{T,NAtype}}, ::Type{S}) where {T,S} = Union{promote_type(T, S),NAtype}
```
causes stack overflow.

nalimilan: Looks like we need to define `Base.promote_rule(::Type{Any}, ::Type) = Any`? Maybe this should be added to Base, and shipped here temporarily with a `VERSION` check?

ararslan:
Bump.

src/abstractdataarray.jl
```diff
@@ -4,7 +4,7 @@
 An `N`-dimensional `AbstractArray` whose entries can take on values of type
 `T` or the value `NA`.
 """
-abstract type AbstractDataArray{T, N} <: AbstractArray{T, N} end
+abstract type AbstractDataArray{T, N} <: AbstractArray{Union{T, NAtype}, N} end
```
rofinn: Should we maybe add an alias (e.g., `const Data{T} = Union{T, NAtype}`) and not export it? Reading code that operates on `DataArrays.Data{T}` seems more readable to me and it might help with deprecation later on.

andreasnoack:
Updated. The test failure on master is most likely because of the macrocalypse going on on master right now. Hopefully, it'll either be fixed or will betold how to fix `DataArrays` in a couple of days.

andreasnoack:
I think this is ready but right now it causes [a segfault](https://github.com/JuliaLang/julia/issues/23351) on master

rofinn:
Hooray, thanks @andreasnoack!

PR #288, opened 2017-09-27
Port to Nulls.jl

nalimilan:
This replaces `NA` with `Nulls.null` and `NAtype` with `Nulls.Null`.
The only breaking change is that `null == null` gives `true`, while `NA == NA`gave `NA` (same for other comparison operators).

-----

The goal of this PR is to make DataArrays compatible with the new Nulls-based DataFrames, so that they can continue to be used until `Array{Union{T, Null}}` is efficient enough, as discussed in https://github.com/JuliaData/DataFrames.jl/pull/1209. The port should be relatively easy for users, since everything goes though relatively simple deprecations.

The most/only breaking change concerns comparison operators (as noted above). However it shouldn't be too painful either since in general it means that code which would previously fail in the presence of `NA`s will now work. The only common situation in which there will be a difference is when user code replaced `NA`s resulting from a comparison manually. Nevertheless, I think we should discuss (in Nulls rather than here) these semantics carefully before committing to them and forcing users to adapt their code.

Tests will fail until we have a release of Nulls.jl including https://github.com/JuliaData/Nulls.jl/pull/30, https://github.com/JuliaData/Nulls.jl/pull/31 and https://github.com/JuliaData/Nulls.jl/pull/32.


src/abstractdataarray.jl
```diff
@@ -20,45 +20,43 @@ A 2-dimensional [`AbstractDataArray`](@ref) with element type `T`.
...
 # Generic iteration over AbstractDataArray's
 
 Base.start(x::AbstractDataArray) = 1
 Base.next(x::AbstractDataArray, state::Integer) = (x[state], state + 1)
 Base.done(x::AbstractDataArray, state::Integer) = state > length(x)
 
-Base.broadcast{T}(::typeof(isna), a::AbstractArray{T}) =
-    NAtype <: T ? BitArray(map(x->isa(x, NAtype), a)) : falses(size(a)) # -> BitArray
-
+# FIXME: type piracy
```
nalimilan: I couldn't find a way to remove this without hurting performance, and since the point of porting DataArrays is to preserve performance... Maybe it's OK to keep this for now.

Another solution would be to move this definition to Nulls.jl, and only implement here the `DataArray` method. But I'm not sure we want to support this kind of API in the long term.

test/operators.jl
```diff
@@ -48,41 +48,49 @@ end
...
 
-    # All unary operators return NA when evaluating NA
+    # All unary operators return null when evaluating null
     for f in [+, -]
-        @test isna(f(NA))
+        @test isnull(f(null))
     end
 
-    # All elementary functions return NA when evaluating NA
+    # All elementary functions return null when evaluating null
```
nalimilan: This is the most significant behavior change. See also below.

nalimilan:
> The most/only breaking change concerns comparison operators (as noted above). However it shouldn't be too painful either since in general it means that code which would previously fail in the presence of NAs will now work. The only common situation in which there will be a difference is when user code replaced NAs resulting from a comparison manually. Nevertheless, I think we should discuss (in Nulls rather than here) these semantics carefully before committing to them and forcing users to adapt their code.

Turns out we'd better use the current `NA` behavior for `null`. See https://github.com/JuliaData/Nulls.jl/pull/33.

nalimilan:
I've added a commit to keep the current logic in which `NA == NA -> NA`, since JuliaData/Nulls.jl#33 just implemented `null == null -> null`.

test/nas.jl
```diff
@@ -1,75 +1,66 @@
...
 
-    dv[[1, 2, end]] = NA
+    dv[[1, 2, end]] = null
 
     a = dropna(dv)
-    @test_throws NAException for v in each_failna(dv); end
-    @test collect(each_dropna(dv)) == a
-    @test collect(each_replacena(dv, 4)) == [4, 4, a..., 4]
-
-    @testset "promotion" for (T1, T2) in ((Int, Float64),
```
nalimilan: I removed these because I couldn't get them to work in the general case, and these should live in Nulls anyway. Cf. https://github.com/JuliaData/Nulls.jl/issues/23 and https://github.com/JuliaStats/DataArrays.jl/pull/280#discussion_r130244769.

nalimilan:
@andreasnoack Opinions? I've tried a few operations mixing DataArrays from this PR and DataFrames master, and everything seems to work fine.

andreasnoack:
At this point, this is just a renaming and migration out the package, right? No semantical changes. Then it should be fine.

nalimilan:
Yes, mostly. The only significant change is the `promote` issue, but I think it's OK now (see comment above and its links). There's also the probable removal of iteration, but that's really small (https://github.com/JuliaData/Nulls.jl/issues/39). Finally, comparison operators have been made consistent with `NA`, so I should be able to reinstate part of the removed tests.

More systematic testing with DataFrames master wouldn't hurt though (once this PR has been merged).

andreasnoack:
Hm. Do tests pass locally? I'd expect them to fail without the promotion definitions. Also, do you need to add `Nulls` to `REQUIRE`?

nalimilan:
Yes, they pass locally, but you need Nulls.jl master. I hope we can tag a release soon.

andreasnoack:
You'll also need to add it to `REQUIRE`. See https://github.com/JuliaStats/DataArrays.jl/blob/9169e9f0a2bfab04cc115428e1c19aa513a9455e/REQUIRE`. That is the current reason why the tests fail.

andreasnoack:
Why can't we make a new release of `Nulls` right now?

nalimilan:
I'm working on it. The problem is that it breaks DataStreams and CategoricalArrays. So better tag compatible releases first (there's one for the former, I'm preparing one for the latter).

nalimilan:
Tests passed on 0.6 now that Nulls 0.1.0 has been tagged. Good to merge?

andreasnoack:
Do you know why it is failing on nightly? It looks related to this change.

nalimilan:
Actually I hadn't even looked at nightlies since they are so broken. The package doesn't load in a more recent version. But looks like as easy fix, so let's try while it's fresh in my mind. I've added a commit.

src/operators.jl
```diff
@@ -213,7 +208,7 @@ end
 # Treat ctranspose and * in a special way
 for (f, elf) in ((:(Base.ctranspose), :conj), (:(Base.transpose), :identity))
     @eval begin
-        $(f)(::NAtype) = NA
+        $(f)(::Null) = null
```
nalimilan: Woops, I've missed this one apparently.

nalimilan:
Nighties have been updated, so now we get a crash there. :-/ That's an "Illegal instruction" just like in DataFrames master, so probably not worth worrying about for now.

I've also removed a few lifted methods which should be defined in Nulls instead, see https://github.com/JuliaData/Nulls.jl/pull/42. That means we'll have to tag another Nulls release...

nalimilan:
Now that Nulls 0.1.1 has been tagged and that CI passes on 0.6, should be ready to merge?

docs/src/da.md
```diff
@@ -5,8 +5,8 @@ CurrentModule = DataArrays
 ```
 
 ```@docs
-NA
-NAtype
+null
```
ararslan: Shouldn't we leave this for the Nulls.jl docs?

nalimilan: Yeah, that's always hard to decide. I've added a commit removing these objects.

ararslan: We could link to the Nulls docs saying something like "DataArrays uses Nulls to represent missing data. For more information about Nulls, see [the Nulls docs]&#8203;(link)."

nalimilan:
I've also reverted the deprecation of `dropna(x)` in favor of `collect(Nulls.skip(x))`, as the performance impact is going to be severe. So I've just renamed it to `dropnull` for now (like `padna` -> `padnull`). We can always revisit this later.

docs/src/da.md
nalimilan: There's already something on the homepage of the manual.

ararslan: Oh sorry, missed that

nalimilan:
After doing a bit more cleanup and adding another deprecation, tests from current master are passing with the implementation from that PR (with a handful of changes were needed). So breakage should be minimal. The only functions I removed without a deprecation are those which would be type piracy: `length`, `size`, `SpecialFunctions.erf`, `SpecialFunctions.erfc`, `SpecialFunctions.digamma`. The cause of the error should be quite easy to identify anyway.

I haven't renamed the `na` field of `DataArray`, because it would be tedious to do for no real gain, and it could break code relying on the internals. Does that make sense?

quinnj:
Sounds good. What were the problematic `length` `size` definitions?

nalimilan:
`length(::Null) = 1` and `size(::Null) = ()`. That is, the ones we just removed from Nulls.jl.

quinnj:
Ah, got it.

src/dataarray.jl
```diff
@@ -5,48 +5,64 @@
...
             throw(ArgumentError(msg))
         end
-        # additionally check that d does not contain NA entries
-        if eltype(d) === Any
-            for i in eachindex(d)
-                if isassigned(d, i) && isna(d, i)
-                    m[i] = true
+        # if input array can contain null values, we need to mark corresponding entries as missing
+        if eltype(d) >: Null
+            # If the original eltype is wider than the target eltype T, conversion may fail
```
nalimilan: @quinnj This is a situation where it would be useful to be able to access the underlying data of a `Union{T, Null}` array, unsafely reinterpreting it as an `Array{T}`. Since `DataArray` stores the missingness mask separately, we are certain we won't access elements which are `null`. Is there a hacky way to do that currently? Would it make sense to provide an API to do that?

quinnj: Yeah, we need to figure out the right api for that. It was in the original isbits Union optimizations PR, but @vtjnash and @yuyichao had concerns w/ how it was implemented, so I just took it out for the initial PR. I'll open an issue so we can push forward on it now that everything's in Base.

vtjnash: I think that's just a type-assert (using the eltype parameter from DataArrays)?

nalimilan: But how can I convert an `Array{Union{T, Null}}` to an `Array{T}` without copying in the presence of `null` entries?

vtjnash: Seems like something reinterpret should be allowed to do

nalimilan: But it returns a `ReinterpretArray` now, so that would require allowing the field to be any `AbstractArray, right? Not the end of the world, probably.

quinnj:
how are things @nalimilan looking?

nalimilan:
I think that's mostly good. I'd like to define `levels` in Nulls (https://github.com/JuliaData/Nulls.jl/pull/46) and override it here, so that DataArrays and CategoricalArrays can be used at the same time without conflicts.

I'm also checking that old DataArrays-based DataFrames tests pass with the new Nulls-based DataFrames (modulo a few reasonable changes where needed). So far it mostly works, but that process allowed me to catch some bugs and problematic changes (like https://github.com/JuliaData/DataFrames.jl/issues/1249), so I'd rather finish it before merging this PR.

quinnj:
Sounds great; thanks for doing all this.

nalimilan:
I think this should be ready for a final review. With this branch, DataFrames tests with `DataArray` columns pass (https://github.com/JuliaData/DataFrames.jl/pull/1260). We should also deprecate `PooledDataArray` in favor of `CategoricalArray` (or `PooledArray` depending on the case), but we can do that later.

src/dataarray.jl
```diff
@@ -377,29 +393,7 @@ function Base.unique(da::DataArray{T}) where T # -> DataVector{T}
...
-```jldoctest
-julia> levels(@data [1, 2, NA])
-2-element DataArrays.DataArray{Int64,1}:
- 1
- 2
-```
-"""
-function levels(da::DataArray) # -> DataVector{T}
-    unique_values, firstna = finduniques(da)
+function Nulls.levels(da::DataArray) # -> DataVector{T}
```
quinnj: I'm guess the old docs here got moved somewhere else? Or they just use the docs in Nulls.jl?

src/operators.jl
```diff
@@ -413,85 +376,39 @@ function (==)(a::DataArray, b::DataArray)
...
-        else
-            a[i] == b[i] || return false
-        end
-    end
-    has_na ? NA : true
+    has_null ? null : true
 end
 
 # ambiguity
-@swappable (==)(::NAtype, ::WeakRef) = NA
```
quinnj: I don't think we have this in Nulls, I'm guess it's not that important?

src/operators.jl
```diff
@@ -593,17 +505,10 @@ for f in (:(+), :(-))
 end
 
 # / is defined separately since it is not swappable
-(/)(b::AbstractArray{T,N}, ::NAtype) where {T,N} =
+(/)(b::AbstractArray{T,N}, ::Null) where {T,N} =
     DataArray(Array{T,N}(size(b)), trues(size(b)))
 @dataarray_binary_scalar(/, /, nothing, false)
 
-for f in [:(Base.maximum), :(Base.minimum)]
```
quinnj: These seem like weird definitions on scalars anyway

src/dataarray.jl
nalimilan: Yes, this method just does the same thing as the fallback in Nulls.

src/operators.jl
nalimilan: In practice I doubt it's useful. These were probably implemented at the time ambiguity warnings were printed when loading the package. Though we could enable ambiguity checks in Nulls and fix this kind of situation (there might be others).

src/operators.jl
nalimilan: Yeah, I think that's because numbers are iterable. We could define these in Nulls since that would be consistent with the goal of accepting `null` everywhere `Number` is, but we can also wait until somebody complains so that we're certain it's useful.

src/DataArrays.jl
```diff
@@ -25,23 +26,10 @@ module DataArrays
...
            getpoolidx,
            gl,
-           head,
-           isna,
-           levels,
-           NA,
-           NAException,
-           NAtype,
-           padna,
+           padnull,
```
nalimilan: I've left this method here, but we could define it in Nulls if that's really useful. No hurry, though.

src/dataarray.jl
nalimilan: @vtjnash Am I right?

PR #289, opened 2017-10-19
Make Each(Drop|Replace|Fail)Null iterators faster

nalimilan:
The compiler previously did not know that the result would never be Null.
This dramatically improves their efficiency.

Also require that the type of replacement matches eltype of iterator
for `EachReplaceNull`, which is much faster. `Nulls.replace()` already
takes care of converting it on construction.

Optimizations need to rely on the particular fields of `DataArray`.
This requires restricting them to `DataArray`, and moving code
to dataarray.jl since the type had not been defined in abstractdataarray.jl.

Rather than adding specialized iterators for `PooledDataArray`, which
is going to be deprecated, rely on the generic iterators provided
by `Nulls`, which are already faster than the current ones.

Cf. https://github.com/JuliaData/Nulls.jl/pull/50

---------
Before (on Julia 0.6.0):
```julia
Main> using Nulls

Main> using DataArrays

Main> using BenchmarkTools

Main> x = DataArray(Vector{Union{Int, Null}}(rand(Int, 100_000)));

Main> x[rand(1:length(x), 10_000)] = null;

Main> skip_iter(x) = sum(Nulls.skip(x))
skip_iter (generic function with 1 method)

Main> @btime skip_iter(x);
  6.988 ms (552024 allocations: 8.42 MiB)

Main> replace_iter(x, y) = sum(Nulls.replace(x, y))
replace_iter (generic function with 1 method)

Main> @btime replace_iter(x, 0);
  31.703 ms (1051004 allocations: 17.56 MiB)

Main> y = DataArray(Vector{Union{Int, Null}}(rand(Int, 100_000)));

Main> fail_iter(x) = sum(Nulls.fail(x))
fail_iter (generic function with 1 method)

Main> @btime fail_iter(y);
  31.586 ms (1098980 allocations: 18.29 MiB)
```

After (on Julia 0.6.0):
```julia
Main> using Nulls

Main> using DataArrays

Main> using BenchmarkTools

Main> x = DataArray(Vector{Union{Int, Null}}(rand(Int, 100_000)));

Main> x[rand(1:length(x), 10_000)] = null;

Main> skip_iter(x) = sum(Nulls.skip(x))
skip_iter (generic function with 1 method)

Main> @btime skip_iter(x);
  516.099 μs (4 allocations: 64 bytes)

Main> replace_iter(x, y) = sum(Nulls.replace(x, y))
replace_iter (generic function with 1 method)

Main> @btime replace_iter(x, 0);
  676.586 μs (4 allocations: 80 bytes)

Main> y = DataArray(Vector{Union{Int, Null}}(rand(Int, 100_000)));

Main> fail_iter(x) = sum(Nulls.fail(x))
fail_iter (generic function with 1 method)

Main> @btime fail_iter(y);
  526.172 μs (4 allocations: 64 bytes)
```


src/abstractdataarray.jl
```diff
@@ -46,58 +46,4 @@ julia> isnull(X, 3)
...
-    EachReplaceNull(da, replacement)
-Base.length(itr::EachReplaceNull) = length(itr.da)
-Base.start(itr::EachReplaceNull) = 1
-Base.done(itr::EachReplaceNull, ind::Integer) = ind > length(itr)
-Base.eltype(itr::EachReplaceNull) = Union{Nulls.T(eltype(itr.da)), typeof(itr.replacement)}
-function Base.next(itr::EachReplaceNull, ind::Integer)
-    item = isnull(itr.da, ind) ? itr.replacement : itr.da[ind]
-    (item, ind + 1)
-end
+Base.isnull(a::AbstractArray{T}, i::Real) where {T} = Null <: T ? isa(a[i], Null) : false # -> Bool
```
ararslan: piratey _vector_version svg

nalimilan:
OK to merge?

PR #293, opened 2017-11-12
Port from Nulls to Missings

nalimilan:
Add error messages to `MissingException`, which requires them.

This depends on https://github.com/JuliaData/Nulls.jl/pull/51 being merged and a the new Missings package being tagged.


nalimilan:
OK, this should be ready to merge.

PR #302, opened 2018-06-18
Require Julia 0.7

nalimilan:


nalimilan:
Funny, I expected that many more deprecations would have to be fixed. So we don't actually need to require 0.7, and things need to be handled manually.

PR #304, opened 2018-08-10
Updates for 1.0

quinnj:
Well, this is as far as I got for 1.0 compat here. The reducedim.jl and broadcast.jl changes are way more involved.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,4 @@
 julia 0.7
-Compat 0.43.0
 Missings 0.2.1
 StatsBase 0.15.0
 Reexport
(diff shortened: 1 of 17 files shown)


nalimilan:
Probably better merge the changes you finished first even if tests don't pass, and leave problematic files out for now?

ararslan:
Can we just let this package die with 1.0 with a clear note in the README?

nalimilan:
Given that Jacob ported a lot of the code, I'd rather merge that. Then we'll see whether somebody is interested in doing the rest of the port or not.

